PR #32, opened 2021-05-04
Use branded types and refacto siren and siret checks

XavierJp:


XavierJp:
- introduce `Siren`or `Siret` (branded types)
- focus checks on Siren/siret in the model layer
- clients layer still use `string` but could also use `Siren`or `Siret`. Client layer assumes siren and siret are valid
- View layer manages the redirection based on the exceptions raised by model layer

PR #153, opened 2022-09-01
feature: add verification of VAT by VIES API

XavierJp:
Add a call to VIES

Add a new TVA section that either say the TVA is validated by VIES either explain it is not but still return the computed one.

@skelz0r so far I have found quite a big number of invalid TVA that's why I still add the unvalid number. Open to discussion

I also created a page about vies but I did not add a vies monitor as the API is not public. I might consider adding a monitor, but no apigouvlink.

Finally I decided to useCache for VIES but also for RNA


XavierJp:
Modif : ajout d'un monitor uptime robot en plus

Haelle:
Possible que ton Ping se fasse ban, ils ont une page de monitoring sinon : https://ec.europa.eu/taxation_customs/vies/#/self-monitoring

XavierJp:
Merci pour l'info de la page de ping, je vais probablement faire une semaine tel quel pour voir comment ca répond (est ce que je me fais ban, combien de temps moyen de réponse etc.)

XavierJp:
Closes #154

PR #186, opened 2022-10-04
Feature: Add fake loading bar

XavierJp:


frontend/src/load-bar.js
```diff
@@ -0,0 +1,27 @@
+const wait = async (timeout) => {
+  return new Promise((resolve, reject) =>
+    window.setTimeout(() => resolve(), timeout)
+  );
+};
+
+window.addEventListener('beforeunload', async () => {
+  const loader = document.createElement('div');
+  loader.style.position = 'absolute';
```
maudetes, line 9: Could we create a css rule for the loader, and only play on width here? :)

frontend/src/load-bar.js
```diff
@@ -0,0 +1,27 @@
...
+  loader.style.left = '0';
+  loader.style.background = '#000091';
+  loader.style.width = '30vw';
+  loader.style.height = '4px';
+  loader.style.zIndex = 1000;
+
+  document.body.appendChild(loader);
+
+  for (let w of [
+    0, 3, 5, 10, 12, 15, 22, 28, 34, 35, 36, 38, 40, 45, 65, 66, 78, 80, 85, 88,
```
maudetes, line 21: How about a random increase? Filtered with a max

maudetes (commented):
Nice UX I think :)

PR #211, opened 2022-12-05
Refacto: move tabs into an object rather than jsx

XavierJp:
Use an object instead of hardcoding tabs in jsx.

It will now be much easier to customize tabs based on uniteLegale type for instance


yoanficadiere:
LGMT !

yoanficadiere (approved):

PR #216, opened 2022-12-06
Revert "Fix rate limiting search api"

XavierJp:
Reverts etalab/annuaire-entreprises-site#215


XavierJp:
Will move this logic in a dedicated `complements` model so this PR is no longer relevant

PR #233, opened 2022-12-12
224 DX

yoanficadiere:
Adding some rules in eslint and use typescript path for import


XavierJp (commented):
First look seems good to me but there is conflict with `main` + could not build in local & could not test. 

We can discuss it tmr

.eslintrc
```diff
@@ -1,13 +1,61 @@
 {
-  "extends": "next",
   "rules": {
+    "no-console": [
+      "error",
+      {
+        "allow": ["warn", "error"]
+      }
+    ],
+    // no need for unique key in maps as we dont use react on client/front
```
XavierJp, line 9: Either remove the comment or it is in the wrong place

cypress/e2e/dynamic-pages.cy.js
```diff
@@ -15,13 +15,6 @@ siret.forEach((siret) => {
         expect(resp.status).to.eq(200);
       });
     });
-
```
XavierJp, line 18: Why delete this test ?

jest.config.js
```diff
@@ -1,5 +1,23 @@
-module.exports = {
+const { pathsToModuleNameMapper } = require('ts-jest');
```
XavierJp, line 1: Are we force to use this ? Feels kinda hacky

XavierJp: But again it is fine. it is only the tests.

XavierJp (approved):

jest.config.js
yoanficadiere: Yeah this dev-dependencie is very usefull to make jest able to supports all features of TypeScript

PR #282, opened 2023-01-09
chore: use regex for commit linting

XavierJp:
This is an attempt at a much much faster and simpler commit lint using bash and regex


XavierJp:
@HAEKADI

yoanficadiere (approved):

PR #301, opened 2023-01-20
chore: delete advanced search js

XavierJp:
Alternative to `react-select` of geo filter. Only apply to Geo filters. React-select seems perfect for NAF/APE selection

Pros :
- easier to fix / tweak
- comply with dsfr and Jeremie's mock up

Cons
- not a lib
- specific to geo search

### DIFF
--- a/components/search-filters/filter-geo.tsx
+++ b/components/search-filters/filter-geo.tsx
@@ -15,7 +15,7 @@ export const FilterGeo: React.FC<{ cp_dep?: string; cp_dep_label?: string }> =
     const [isLoading, setLoading] = useState(false);
     const [geoSuggests, setGeoSuggests] = useState([]);
 
-    const geoSearchRef = useRef(null);
+    const geoSearchRef = useRef<HTMLInputElement | null>(null);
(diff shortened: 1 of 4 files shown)


components/search-filters/filter-geo.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const FilterGeo: React.FC<{ cp_dep?: string; cp_dep_label?: string }> =
+  ({ cp_dep, cp_dep_label }) => {
+    const [labelDep, setLabelDep] = useState(cp_dep_label);
+    const [dep, setDep] = useState(cp_dep);
+    const [isLoading, setLoading] = useState(false);
+    const [geoSuggests, setGeoSuggests] = useState([]);
+
+    const geoSearchRef = useRef(null);
```
destpat, line 18: To avoid ts-ignore line 37 it should be typed 

```suggestion
    const geoSearchRef = useRef<HTMLInputElement | null>(null);
```

destpat (approved):

PR #313, opened 2023-01-25
feat: adding multi select for naf code, and refacto

yoanficadiere:


XavierJp (approved):
Perfect. Nice refacto BTW

PR #321, opened 2023-01-30
feat: adding getter for env variable

yoanficadiere:


env.ts
```diff
@@ -0,0 +1,40 @@
+const getEnvValue = (
+  key: string,
+  value?: string,
+  nullable = false
+): string | undefined => {
+  if (typeof value === 'undefined' && !nullable) {
```
XavierJp, line 6:
```suggestion

const getEnvValue = (
  key: string,
  nullable = false
): string | undefined => {
	const value = process.env[key]
  	if (typeof value === 'undefined' && !nullable) {
  		throw new Error(`Missing env variable: ${key}`);
 	}
 	return value;
 }
 
const nonNullableEnvKeys = [
	'INSEE_CLIENT_ID_FALLBACK'
	...
];

const nullableEnvKeys = [
	'ALTERNATIVE_SEARCH_ROUTE',
	...
]
 
 const initializeEnvConfig = () => {
 	return [
 		nonNullableEnvKeys.map(getEnvValue)
 		nullableEnvKeys.map(k=>getEnvValue(k, true))
 	]
```

I think something like this would be simpler to update when adding a new env var. Just add a key in the relevant list.

.env.example
```diff
@@ -0,0 +1,20 @@
+ALTERNATIVE_SEARCH_ROUTE=
```
XavierJp, line 1: 😅 lol good thing I havent updated secrets yet

XavierJp:
Effectivement ca pète totalement les tests

PR #369, opened 2023-03-02
feat: test cypress split  on ci

yoanficadiere:


XavierJp (approved):
We should remove the Print step and try with 4 or 5 matrix

PR #430, opened 2023-04-06
feat: mswjs and mocks

yoanficadiere:


mocks/handlers/insee/token.ts
```diff
@@ -0,0 +1,6 @@
+export const token = {
```
XavierJp, line 1: why not just returning dummy rathen than actual payload ? As it will never be send to Insee API

mocks/handlers/insee/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import { token } from './token';
+
+export const handlers = [
+  rest.post('https://api.insee.fr/token', (_req, res, ctx) => {
+    return res(ctx.json(token));
+  }),
+  rest.get(
+    `https://api.insee.fr/entreprises/sirene/V3/siren/${mockMapping.rge}`,
+    (_req, res, ctx) => {
+      return res(ctx.json({}));
```
XavierJp, line 13: Weird to return empty json here !

XavierJp (changes requested):

Issue #496, opened 2023-05-04
[BUG] Investigate possible event loop blockage in Node


XavierJp:
https://stackoverflow.com/questions/17960452/how-can-i-get-a-list-of-callbacks-in-the-node-work-queue-or-why-wont-node-ex

See also : 
- https://github.com/mafintosh/why-is-node-running
- https://www.ashbyhq.com/blog/engineering/detecting-event-loop-blockers
- https://github.com/naugtur/blocked-at

```
2023-04-27T16:27:25: status=408 time=857 isFromCached=false handleCount=48 reqCount=994 request=https://api.insee.fr/entreprises/sirene/V3/siren/497502872
2023-04-27T16:27:25: status=408 time=863 isFromCached=false handleCount=48 reqCount=997 request=https://api.insee.fr/entreprises/sirene/V3/siret/
2023-04-27T16:27:25: status=408 time=864 isFromCached=false handleCount=48 reqCount=997 request=https://api.insee.fr/entreprises/sirene/V3/siret/
2023-04-27T16:27:26: status=408 time=857 isFromCached=false handleCount=44 reqCount=994 request=https://api.insee.fr/entreprises/sirene/V3/siren/804594315
2023-04-27T16:27:26: status=408 time=861 isFromCached=false handleCount=44 reqCount=997 request=https://api.insee.fr/entreprises/sirene/V3/siret/
2023-04-27T16:27:26: status=408 time=862 isFromCached=false handleCount=44 reqCount=997 request=https://api.insee.fr/entreprises/sirene/V3/siret/
2023-04-27T16:27:28: status=408 time=5009 isFromCached=false handleCount=45 reqCount=997 request=https://recherche-entreprises.api.gouv.fr/search?per_page=10&page=1&q=904756384&limite_matching_etablissements=3&inclure_slug=true&inclure_etablissements=true
2023-04-27T16:27:29: status=408 time=857 isFromCached=false handleCount=50 reqCount=1002 request=https://api.insee.fr/entreprises/sirene/V3/siren/904756384
2023-04-27T16:27:29: status=408 time=858 isFromCached=false handleCount=50 reqCount=1004 request=https://api.insee.fr/entreprises/sirene/V3/siret/
2023-04-27T16:27:29: status=408 time=858 isFromCached=false handleCount=50 reqCount=1004 request=https://api.insee.fr/entreprises/sirene/V3/siret/
[...]

PR #498, opened 2023-05-04
Improve performance issue

XavierJp:
references #496


XavierJp:
@jdesboeufs very good idea, as redis is only used for cache and TTL ~ 10min.

https://github.com/etalab/annuaire-entreprises-infrastructure/pull/9

yoanficadiere (approved):

PR #537, opened 2023-06-21
fix: nps result

yoanficadiere:
Fixing values in NPS

### DIFF
--- a/clients/matomo/index.ts
+++ b/clients/matomo/index.ts
@@ -59,7 +59,7 @@ const getLabel = (labelAsString: string, index: number) => {
  * Aggregate event by month and userType
  */
 const aggregateEvents = (
-  matomoEventStats: { label: string; nb_events: number }[]
+  matomoEventStats: { label: string; nb_events: number }[][]
 ) => {
   const months: {
     [monthKey: string]: { [userTypeKey: string]: number[] };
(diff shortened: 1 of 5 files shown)


clients/matomo/index.ts
```diff
@@ -329,10 +354,24 @@ const createEventsCategoryUrl = () => {
...
  * Compute matomo API url to extract events count
  */
-const createNpsEventUrl = () => {
+const getNpsEvent = async () => {
   const lastYear = getLastYear();
-  const dateRange = `${YYYYMMDD(lastYear)},${YYYYMMDD(new Date())}`;
-  return routes.matomo.report.npsEvents + dateRange;
+  const subtableIdsResponse = await getEventSubtableIds();
+
+  const data = await Promise.all(
```
XavierJp, line 361: Why do a promise all and not a bulk request here ?

functions/compute-stats.js
```diff
@@ -69,10 +69,10 @@ exports.handle = function (event, context) {
   const categories = `https://stats.data.gouv.fr/index.php?module=API&format=json&idSite=145&period=day&method=Events.getCategory&module=API&showColumns=label&filter_limit=9999&date=${dateAsString}`;
 
   get(categories)
```
XavierJp, line 71: You can actually entirely remove `functions/compute-stats.js` as it is no longer used

XavierJp (approved):

XavierJp (approved):

PR #548, opened 2023-06-28
feat: create match all polyfill to support safari 12

yoanficadiere:


frontend/src/utils.js
```diff
@@ -10,7 +19,7 @@ const extractSirenOrSiretSlugFromUrl = (url) => {
   }
   // match a string that ends with either 9 digit or 14 like a siren or a siret
   // we dont use a $ end match as there might be " or %22 at the end
-  const matches = url.matchAll(/\d{14}|\d{9}/g);
+  const matches = matchAllPolyfill(url, /\d{14}|\d{9}/g);
```
XavierJp, line 22: Reading the code, I am not sure than `matchAll` is needed here.

Goal is to match : 

`skdjfjsdhfsdf999888999dfsf` and to extract `999888999`

XavierJp: I think you can also use https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/match

XavierJp (approved):
Supprimer `matchAllPolyfill` dans `utils/helpers/siren-and-siret.ts` et on est bon

PR #556, opened 2023-07-06
fix: create a dedicated message for 400 bad requests

XavierJp:
- revamp contact form links
- add an improved warning for bad request to avoid confusion and help get bug feedbacks


HAEKADI (approved):
🚀

PR #566, opened 2023-07-20
chore: refacto idcc

XavierJp:
- Use API Recherche Entreprise for Idcc
- Stop using siret2Idcc
- Redirect to Code du travail Numérique


HAEKADI (approved):
🚀

PR #568, opened 2023-07-21
fix: e2e staging recipe

XavierJp:


XavierJp:
@HAEKADI I think deactivating video recording sealed the deal !

HAEKADI (approved):
Niice 🚀

PR #638, opened 2023-09-19
fix: add a specific message in section association when rna api is KO

### LINKED ISSUE #630: [Associations] Faire apparaitre un message d'erreur quand l'API Association ne répond pas
## Problème 

Quand le client de l’API Association est retourne un 408 https://github.com/etalab/annuaire-entreprises-site/blob/main/clients/api-proxy/association/index.ts

Le front ignore l'erreur et affiche un message suggérant que l'Association n'a pas été retrouvée 
https://github.com/etalab/annuaire-entreprises-site/blob/main/models/association.ts#L15
https://github.com/etalab/annuaire-entreprises-site/blob/main/components/association-section/index.tsx

## Proposed solution

Change association model association to either be `IAssociation` or `IAPINotResponding`

Use `isAPINotResponding` to render `<AdministrationNotResponding />` https://github.com/etalab/annuaire-entreprises-site/blob/c7aa4b91deedeeb40230ab5e0bd4309d8032b332/components/administration-not-responding/index.tsx#L11

For implementation example see :
https://github.com/etalab/annuaire-entreprises-site/blob/c7aa4b91deedeeb40230ab5e0bd4309d8032b332/models/api-not-responding.ts#L18
https://github.com/etalab/annuaire-entreprises-site/blob/main/components/finances/societe/index.tsx#L22

johangirod:
- Add a new administration : DJEVPA
- Modify type of association response accordingly
- Change wording for accessing link to api doc

fix #630

### DIFF
--- a/components/api-monitoring/index.tsx
+++ b/components/api-monitoring/index.tsx
@@ -219,7 +219,7 @@ const ApiMonitoring: React.FC<IMonitoringWithMetaData> = ({
             Envie de réutiliser cette API ?{' '}
             {apigouvLink && (
               <a href={apigouvLink} target="_blank" rel="noreferrer noopener">
-                Consulter la fiche api.gouv.fr
+                Consulter la documentation
               </a>
             )}
           </i>
--- a/components-ui/alerts/association-adress.tsx
+++ b/components-ui/alerts/association-adress.tsx
@@ -1,4 +1,5 @@
 import { INSEE, MI } from '#components/administrations';
+import { isAPINotResponding } from '#models/api-not-responding';
 import { IAssociation } from '#models/index';
 import { getAdresseUniteLegale } from '#models/statut-diffusion';
 import { ISession } from '#utils/session';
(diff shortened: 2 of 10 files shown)


components/api-monitoring/index.tsx
```diff
@@ -219,7 +219,7 @@ const ApiMonitoring: React.FC<IMonitoringWithMetaData> = ({
             Envie de réutiliser cette API ?{' '}
             {apigouvLink && (
               <a href={apigouvLink} target="_blank" rel="noreferrer noopener">
-                Consulter la fiche api.gouv.fr
+                Consulter la documentation
```
XavierJp, line 222: Nice extra touch 🙏

XavierJp (changes requested):
Great. A few fix needed : 
- build seems broken. Probably an issue with types
- `isAPINotResponding` unfortunately often is not a sufficient check. It includes both 5xx and 4xx errors. So it include 404 that depending on the API can either be a real error or have a business meaning.

In the case of RNA I think 404 has not a very relevant business meaning but it also is not an error. It only means association was not found in RNA. It does not mean that the association does not exist as RNA cannot be trust.

XavierJp:
If it is ok with you I can push fixes to deploy this today.

XavierJp (approved):

PR #640, opened 2023-09-19
[RNE] Les données RNE des pages /justificatif et /dirigeants sont appelées en async

### LINKED ISSUE #632: [RNE] Faire les appels au RNE des pages /justificatif et /dirigeants depuis le browser

johangirod:
fix #632


XavierJp:
See #646 

Also we need to think of a strategy to avoid having anyone calling this API route. As they might overflow us. 

I think we shoould by default create a session for every visitor. With no specific right. And then only allow these routes for our sessions

components/annonces-section/bodacc.tsx
```diff
@@ -45,6 +45,7 @@ const AnnoncesBodaccSection: React.FC<{
     return (
       <LoadingSection
         title="Annonces BODACC"
+        description="Nous récupérons les annonces de cette structure"
         sources={[EAdministration.DILA]}
       />
     );
```
XavierJp, line 51: To add an animation, it would need to merge `<LoadingSection />` into the default `<Section />` component

components/dirigeants-section/beneficiaires.tsx
```diff
@@ -4,19 +4,19 @@ import InpiPartiallyDownWarning from '#components-ui/alerts/inpi-partially-down'
...
 } from '#models/api-not-responding';
 import { IBeneficiaire, IImmatriculationRNE } from '#models/immatriculation';
-import { formatDate, formatDatePartial } from '#utils/helpers';
-import { Siren } from '#utils/helpers';
+import { Siren, formatDatePartial } from '#utils/helpers';
 import AdministrationNotResponding from '../administration-not-responding';
 
 type IProps = {
-  immatriculationRNE: IImmatriculationRNE | IAPINotRespondingError;
+  immatriculationRNE: IImmatriculationRNE | IAPINotRespondingError | null;
```
XavierJp, line 19: How could we use a cleaner way to materialize the loading state ? 

For instance use a dedicated type like `IAPILoading`rather than `null` with its dedicated type check `isAPILoading` ?

models/api-not-responding.ts
```diff
@@ -15,11 +15,11 @@ export const APINotRespondingFactory = (
...
-): toBeDetermined is IAPINotRespondingError => {
+export function isAPINotResponding<T>(
+  toBeDetermined: Exclude<T, null> | IAPINotRespondingError
+): toBeDetermined is IAPINotRespondingError {
   if ((toBeDetermined as IAPINotRespondingError).errorType) {
     return true;
   }
   return false;
-};
+}
```
XavierJp, line 25: Very nice !

XavierJp: Note for future, we should also use this on `httpClient`

clients/api-proxy/rne/index.ts
```diff
@@ -11,6 +11,6 @@ import { clientAPIProxy } from '../client';
 export const fetchRNEImmatriculation = async (siren: Siren, useCache = true) =>
   clientAPIProxy(
     routes.proxy.rne + siren,
-    { timeout: constants.timeout.XL },
+    { timeout: constants.timeout.XXL },
```
XavierJp, line 14: XXL = 20s

I think we can create `XXXL` for in-browser calls ! (30 sec ? Maybe more depending on what is the default browser timeout)

XavierJp (approved):
Very nice. As we move more and more call to the browsers, I think that we should :
- add the animation as it will pave the way for a default pattern for browser calls
- type the loading/isLoading state as relying on `null` is clunky

clients/api-proxy/rne/index.ts
XavierJp: Actually I am not even sure there is a timeout for ajax calls

johangirod: > Actually I am not even sure there is a timeout for ajax calls

I don't think there is, or it's browser-based. However, the value here is the timeout for the `ProxyAPI` client, which is used in a `/api` route of the next app.

XavierJp: 👍
I created a dedicated issue 
https://github.com/etalab/annuaire-entreprises-api-proxy/issues/17

XavierJp (commented):
I feel like we went down a quite complicated road : 

A lot of logic is in `<APISection />` would be simpler if in the parent component 

Exemple : loading, plural etc.

Could we maybe have APISection be a very basic presentation componenet that can have four states and do only presentations / animations ?

Like : 
1. Loading
2. Data
3. Error
4. Hiding(not applicable)

And the states would be determined by the parent component and the props

XavierJp (commented):
The more I read this PR the more I like this new architecture. 

I am not 100% sure about my changes proposal. This commit :
- renamed a bunch of stuff, especially `APISection` into `DataSection` and `APIRequest` into `data`. I am really not sure about this
- refacto `APISection` code in an attempt to differentiate where the animation gets called and where the business logic is
- refacto `APISection` signature to look more like `Section` and ease future migration
- refacto `AdministrationNotResponding` to ease migration. Target is to entirly move to `DataSection` and not directly use AdministrationNotResponding anymore

XavierJp:
We might also need a `hideIfLoading` behaviour

components/section/data-section.tsx
```diff
@@ -0,0 +1,71 @@
...
+import DataSectionLoader from './data-section-loader';
+
+interface IDataSectionProps<T> extends ISectionProps {
+  data: IAPINotRespondingError | IAPILoading | T;
+  hideIf404?: boolean;
+  additionalInformationOnError?: React.ReactNode;
+  children: (data: T) => JSX.Element;
+}
+
+export function SectionStateMachine<T extends {}>({
```
johangirod, line 20: Tant qu'il n'est pas utilisé, on peut éviter de l'exporter à mon avis ?

components/section/data-section.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const lastModified = data?.lastModified || null;
+
+  if (isAPINotResponding(data) && data.errorType === 404 && props.hideIf404) {
+    return null;
+  }
+
+  return (
+    <Section {...props}>
+      <HeightTransition>
+        <FadeIn>
```
johangirod, line 67: Le `FadeIn` devrait plutôt entourer le `chidren` de `SectionStateMachine`, afin d'être activé lorsque le résultat apparaît, et non au début. L'autre solution est d'ajouter data en `key`, pour effectuer un rerendu lorsque l'état change.

hooks/use-immatriculation-RNE.ts
```diff
@@ -13,10 +17,16 @@ export const useImmmatriculationRNE = (uniteLegale: IUniteLegale) => {
...
-        RNE_ROUTE + '/' + uniteLegale.siren
-      );
-      setImmatriculationRNE(response);
+      try {
+        const response = await httpFrontClient<IImmatriculationRNE>(
+          RNE_ROUTE + '/' + uniteLegale.siren
+        );
+        setImmatriculationRNE(response);
+      } catch {
+        setImmatriculationRNE(
```
johangirod, line 26: :+1: Bien vu ! Dans le futur, on pourrait factoriser cette logique ?

components/immatriculations/rne.tsx
```diff
@@ -22,59 +22,71 @@ type IProps = {
...
   return (
     <>
-      <APISection
+      <DataSection
         id="rne"
         title="Inscription au RNE"
         sources={[EAdministration.INPI]}
-        APIRequest={immatriculation}
+        data={immatriculation}
         hideIf404
```
johangirod, line 87: Après avoir étudié un peu les utilisations de `AdministrationNotResponding`, si on souhaite systématiser le recours à `SectionData`, il manque à pouvoir spécifier une vue spécifique en cas de 404. Plutôt que `hideIf404` on pourrait plutôt avoir `notFoundInfo`, qui peut être à `null`.

johangirod (commented):
C'est top, j'aime beaucoup la refacto ! Ça me semble bien plus clair. On perd la logique du pluriel / singuler dans les titres, mais c'est très facile à réimplémenter, quitte à avoir un petit helper pour cela.

clients/open-data-soft/clients/bodacc.ts
```diff
@@ -44,7 +44,11 @@ interface IBodaccB extends IBodaccCoreRecord {
   radiationaurcs?: string; // "{\"radiationPM\": \"O\"}"
 }
 
+async function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
 const clientBodacc = async (siren: Siren): Promise<IAnnoncesBodacc> => {
+  await sleep(1000);
```
XavierJp: sleep should be removed

components/dirigeants-section/summary.tsx
```diff
@@ -53,8 +57,10 @@ const DirigeantSummary: React.FC<IDirigeants> = ({
   if (inseeDirigeant) {
     summaries.push(<a href="#insee-dirigeant">{inseeDirigeant}</a>);
   }
-
-  if (!isAPINotResponding(immatriculationRNE)) {
+  if (
+    !isAPILoading(immatriculationRNE) &&
+    !isAPINotResponding(immatriculationRNE)
+  ) {
```
XavierJp, line 63: Would be nice to work on `<DataSummary />` one day as it is a very similar logic to `<DataSection />`

Maybe in #667

XavierJp (approved):
- Weird that `/section/data-source` is empty
- did not understand the mix between react transition and css transition
- really like the changes

A few stuffs to clean, and lets ship it !

PR #646, opened 2023-09-22
feat: move bodacc to the front

XavierJp:
This : 
- moves bodacc to the front
- creates new `httpFrontClient` for front use (wraps `fetch`)
- creates a new `odsFrontClient` for frontend call to ODS

Also recent changes to linting and import order create a lot of noise. Sorry :)


models/annonces/bodacc.ts
```diff
@@ -0,0 +1,24 @@
...
+import { verifySiren } from '#utils/helpers';
+import logErrorInSentry from '#utils/sentry';
+import { IAnnoncesBodacc } from '.';
+
+export const getAnnoncesBodaccFromSlug = async (
+  slug: string
+): Promise<IAnnoncesBodacc | IAPINotRespondingError> => {
+  const siren = verifySiren(slug);
+  try {
+    return await clientBodacc(siren);
```
johangirod, line 16: It's a little annoying that the models hold the logic of choosing between server / browser HTTP client. It means that we ough to know if the model is node or browser compatible when we call it. We might want to inject a client or rewrite it to be isomorphic (and choose between fetch and axio depending on a env variable for instance ?).

XavierJp: Definitely agree with you !

I think the env var is hard to implement as we might need either axios or fetch depending on the API we call

I like the idea of injecting the client but i also believe the frontend should not "know" about internal mechanisms of the client layer. Front could also pass an argument to the model that will again pass it to the client layer that would then choose between axios and fetch. But still a bit complicated.

Or we can wrap both axios and fetch into a uniq client wrapper that test something like `typeof window === undefined` to choose between the two clients.

This need some rewrite in order to have the same signature for both clients but i like how it keeps the logic inside the client layer.

XavierJp: See #647

XavierJp: See #648

PR #650, opened 2023-09-25
feat(tech): protect API route exposed on server

johangirod:
By preventing them to be called outside browser navigation

We use a session UUID, valid during the entire session TTL. By default, it's 14 days. If the cookie isn't present in the request, the protected route call throw a 401. Every HTML route in /page will start a session if none exists.

**Drawbacks** 

This solution is imperfect, as it's straightforward to bypass : you just need to call a HTML page (like /), get the cookie, and use it in your API request without limit during the next 14 days.

The problem is that we cannot shorten the TTL of the session, since it's the same object used for auth as an agent.

**Futur improvement** 

We might want to couple this with rate limiting. We can use the session ID for as a key for rate limiting. However, it can be easy to overcome with a pool of session call sequentially... Maybe the solution is to use IP as well ?

### Todo
- [x] rename withVerifySessionApiRoute to withAntiBot
- [ ] remove uuid
- [ ] add timestamp logic
- [x] add sentry log
- [x] error message should be less explicit (keep only the status ?)

### DIFF
--- a/utils/server-side-props-helper/error-handler.ts
+++ b/utils/server-side-props-helper/error-handler.ts
@@ -58,7 +58,7 @@ const handleExceptions = (exception: any, req: IncomingMessage | undefined) => {
   }
 };
 
-export const getScope = (req: IncomingMessage | undefined, slug: string) => {
+const getScope = (req: IncomingMessage | undefined, slug: string) => {
   let sirenOrSiret = {} as any;
 
   try {
--- a/utils/server-side-props-helper/error-handler.ts
+++ b/utils/server-side-props-helper/error-handler.ts
@@ -58,7 +58,7 @@ const handleExceptions = (exception: any, req: IncomingMessage | undefined) => {
   }
 };
 
-const getScope = (req: IncomingMessage | undefined, slug: string) => {
+export const getScope = (req: IncomingMessage | undefined, slug: string) => {
   let sirenOrSiret = {} as any;
 
   try {
--- a/utils/session/with-anti-bot.ts
+++ b/utils/session/with-anti-bot.ts
@@ -2,7 +2,7 @@ import { withIronSessionApiRoute } from 'iron-session/next';
 import { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
 import { logWarningInSentry } from '#utils/sentry';
 import { getScope } from '#utils/server-side-props-helper/error-handler';
-import { hasSessionId, sessionOptions } from '.';
+import { ISession, sessionOptions } from '.';
 
 /**
  * withAntiBot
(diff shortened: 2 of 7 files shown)


utils/session/with-verify-session-api-route.ts
```diff
@@ -0,0 +1,28 @@
...
+ * the handler.
+ *
+ * @param handler
+ * @returns
+ */
+
+export function withVerifySessionApiRoute(handler: NextApiHandler) {
+  function verifySession(req: NextApiRequest, res: NextApiResponse) {
+    if (!hasSessionId(req.session)) {
+      res.status(401).json({ message: 'Session not started' });
```
XavierJp, line 20: Error message might be too explicit ? I fear a scrapper would immediately know what to look for.

Also I would add a warning in Sentry

XavierJp (approved):
Very nice. I wonder if the uuid is even necessary as I suspect that `iron-session` creates a session when we call it in `post-server-side-props`

So we could just ban every call without a session.

Also this might be a good usecase for Next middlewares : 
https://nextjs.org/docs/app/building-your-application/routing/middleware

Regarding the 14days TTL, do you think we should shorten it ?

utils/server-side-props-helper/error-handler.ts
```diff
@@ -58,7 +58,7 @@ const handleExceptions = (exception: any, req: IncomingMessage | undefined) => {
   }
 };
 
-const getScope = (req: IncomingMessage | undefined, slug: string) => {
+export const getScope = (req: IncomingMessage | undefined, slug: string) => {
   let sirenOrSiret = {} as any;
```
XavierJp, line 62: I dont think we need to export scope

utils/session/with-anti-bot.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param handler
+ * @returns
+ */
+
+export function withAntiBot(handler: NextApiHandler) {
+  async function verifySession(req: NextApiRequest, res: NextApiResponse) {
+    if (!userVisitedAPageRecently(req.session)) {
+      const scope = getScope(req, '');
+      logWarningInSentry('Antibot activated for API route', scope);
+
```
XavierJp, line 24: I would not use getScope here, but rather 
`{ details: req.url }`

utils/session/with-anti-bot.ts
```diff
@@ -0,0 +1,43 @@
...
+function userVisitedAPageRecently(session: ISession | null) {
+  if (!session?.lastVisitTimestamp) {
+    return false;
+  }
+  const now = new Date();
+  const lastVisit = new Date(session.lastVisitTimestamp);
+  const diff = now.getTime() - lastVisit.getTime();
+  return diff < 1000 * 60 * 5; // 5 minutes
+}
```
XavierJp, line 41: I love it !

XavierJp (approved):

XavierJp (approved):

PR #672, opened 2023-10-10
chore: add sentry logs for failed api call in front

johangirod:


hooks/use-immatriculation-RNE.ts
```diff
@@ -22,7 +23,8 @@ export const useImmmatriculationRNE = (uniteLegale: IUniteLegale) => {
           RNE_ROUTE + '/' + uniteLegale.siren
         );
         setImmatriculationRNE(response);
-      } catch {
+      } catch (e) {
+        Sentry.captureException(e);
```
XavierJp, line 27: I guess on this code you can use `logErrorInSentry` helper

XavierJp (approved):
We currently call sentry here on frontend https://github.com/etalab/annuaire-entreprises-site/blob/main/frontend/src/sentry.js

Is Sentry.captureException the best practice ? I sometime just throw the error or use a console.error but this seem cleaner

johangirod:
From what I get, Sentry.init will initialize Sentry to log all uncaught exceptions / error. When you catch an exception, it won't bubble thought, and so will not appear in sentry. 
https://docsbot.ai/chat/skFEy0qDC01GrRrZ7Crs/EPqsd8nu2XmKzWnd45tL

If you call throw, then your error will bubble though the app, which might not be what we want (it could be caught by React Error component). I'm not sure if `console.error` is logged to sentry automatically. Even if it's the case, you'll lose the stacktrace.

XavierJp:
Ok so for the `/frontend` let's use `Sentry.captureException(e)`;

Can you maybe please add it to https://github.com/etalab/annuaire-entreprises-site/blob/795206fe98603f6afce6faa638592734e828738d/frontend/src/inpi-pdf-download.js#L16-L17

I also think that this file is not used anymore and can be removed : https://github.com/etalab/annuaire-entreprises-site/blob/795206fe98603f6afce6faa638592734e828738d/frontend/src/sentry.min.js

And on the react app (the rest of the app) let's use logErrorInSentry (that is a wrapper around `Sentry.captureException(e)`

XavierJp (approved):

PR #688, opened 2023-10-16
refactor: use DataSection instead of Section

### LINKED ISSUE #667: [Tech] Refacto `<Section/>`into `<DataSection />`
Refacto legacy `<Section />` that are using `isAPILoading` into `<DataSection />`

Create a new `<DataSummary />` for page's summary

johangirod:
Whenever its possible.

fix #667

**Bonus** 
Add test for JOAFE and DCA section in association + API client as well

**Questions** 
- I don't really get the proposition of creating a <DataSummary> component. What would be its purpose ?
- Should I continue to increase test coverage in this PR ? I think it's a good opportunity to test for all `<DataSection/>` in cypress and their corresponding APIs with snapshot testing.


XavierJp:
Maybe we can create a dedicated PR for further E2E tests as this PR is already pretty big

components/association-section/index.tsx
```diff
@@ -36,33 +31,69 @@ const AssociationNotFound: React.FC<{
...
-    return (
-      <AdministrationNotResponding
-        administration={data.administration}
-        errorType={data.errorType}
-      />
-    );
+    // This case should not happen : if AssociationSection is rendered, it means
+    // that the uniteLegale is an association, and so data is not null (though it
+    // can be of type IAPINotRespondingError).
+    return null;
```
XavierJp, line 49: Not sure to understand why `<AssociationNotFound uniteLegale={uniteLegale} />;`became `null` ?

components/immatriculations/joafe.tsx
```diff
@@ -3,75 +3,64 @@ import ButtonLink from '#components-ui/button';
...
-
-  const data = [
-    ['Siren', formatIntFr(immatriculation.siren)],
-    ['N°RNA', formatIntFr(immatriculation.idRna)],
-    ['Date d’enregistrement', formatDate(immatriculation.datePublication)],
-  ];
-
-  return (
-    <>
-      {immatriculation.downloadLink && (
```
johangirod, line 43: @XavierJp 
Changement de comportement ici également : on affiche la section même s'il n'y a pas de downloadlink. Par contre, on n'affichera pas de lien pour le téléchargement dans ce cas.

johangirod:
@XavierJp J'ai enlevé les tests de cette PR pour la rendre plus facile à review, je vais en créer une nouvelle pour les tests.

components/association-section/index.tsx
```diff
@@ -36,33 +32,77 @@ const AssociationNotFound: React.FC<{
...
     return (
-      <AdministrationNotResponding
-        administration={data.administration}
-        errorType={data.errorType}
-      />
+      <Section
+        title={`Répertoire National des Associations`}
+        sources={[EAdministration.MI]}
+      >
+        <AssociationNotFound uniteLegale={uniteLegale} />;
```
XavierJp, line 55: There is an issue here <AssociationNotFound /> needs to be rewritten as it already contains a Section https://github.com/etalab/annuaire-entreprises-site/blob/e69d27d9f87bd3040b4674ef3c731354107699c1/components/association-section/index.tsx#L15

components/immatriculations/joafe.tsx
XavierJp: Sounds good

components/association-section/index.tsx
johangirod: My bad.

johangirod: Not anymore, I removed it in this PR : https://github.com/etalab/annuaire-entreprises-site/pull/688/files/57481e5c6bc07d13e41e7947e80b4009d6ab0591#diff-b322bb495d9629eff3b2790166611b51f0402a9edceb3768813c053eb17ecb21L42

Indeed, it is used in `notFoundInfo` of `DataSection` props which wraps it inside `Section` when displaying it.

PR #692, opened 2023-10-17
fix: change cron time and remove duplicate notify

XavierJp:
- change cron -> github is in GMT
- remove duplicated notify


XavierJp:
Actually duplicated has already been removed by #687

PR #711, opened 2023-10-23
chore: improve compat for older browser

johangirod:
***Investigating***


XavierJp:
https://github.com/etalab/annuaire-entreprises-site/blob/main/utils/server-side-props-helper/post-server-side-props.ts#L20

XavierJp:
Is it still relevant or can we close it ? #762

PR #725, opened 2023-10-25
Table of Content in definition page

johangirod:
- refactor faq and definitions to use the same basetype and `loadAll` function
- parse markdown statically instead of using react-markdown
- automatically add slug id to headings, retrieve headings info and add them to `IArticle` type
- add toc in definition pages
- remove `react-markdown` 
- add heading anchor link 
- Add a `summary` ui component


pages/definitions/[slug].tsx
```diff
@@ -1,56 +1,72 @@
...
+                  <a href={`#${id}`}>{value}</a>
+                </li>
+              ))}
           </ul>
-        </div>
-      ) : null}
-    </TextWrapper>
-  </>
-);
+        </nav>
```
Comment (XavierJp) [inline pages/definitions/[slug].tsx:43]: Might be worth it to  move it into a `ui-component`

pages/definitions/[slug].tsx
```diff
@@ -1,56 +1,72 @@
...
+
+        <div dangerouslySetInnerHTML={{ __html: definition.body.html }} />
```
Comment (XavierJp) [inline pages/definitions/[slug].tsx:45]: Also worth it to  move it into a `ui-component` as it will replace every former `<ReactMarkdown />`

utils/static-markdown-pages/load-all.ts
```diff
@@ -0,0 +1,56 @@
...
+    raw: body,
+    html: parsedBody.value as string,
+    // @ts-ignore
+    headings: parsedBody.data.headings.map((header) => ({
+      id: header.data.id,
+      value: header.value,
+      depth: header.depth,
+    })),
+  };
+}
```
XavierJp, line 49: I really like it. If I may suggest : we could have a dedicated type 
```
{
    raw: string;
    html: string;
    headings: { id: string; value: string; depth: number }[];
  }
```

And every parsed object such as IArticle could extends that type

XavierJp (commented):
Really like what you did with the merging of FAQ and definition 👍 

Plus I think we need to push this PR all the way and remove entirely `react-markdown` (it is not used that much accross the app) and we should totally remove unified if possible

Be careful as I used some of react markdown remapping capabilites for formatting purpose : https://github.com/etalab/annuaire-entreprises-site/blob/b6306fcee2b2495f6f7b49c43ad2b7fba666d605/pages/historique-des-modifications.tsx#L29-L30

XavierJp:
We should  create some `IMarkdown` types useful to type `markdownProcessor` signature and for your `props`

XavierJp:
Plus one thing that would really be lovely would be to add visual anchors like I did in https://api.gouv.fr/les-api/api-particulier#a-quoi-sert-l%E2%80%99api-particulier-?

I always felt contrained by react markdown so I did not do it but I find these anchor very helpful for support

pages/definitions/[slug].tsx
Comment (johangirod) [reply pages/definitions/[slug].tsx:43]: Yes, I need to create a new Sommaire component
https://www.systeme-de-design.gouv.fr/elements-d-interface/composants/sommaire

pages/definitions/[slug].tsx
Comment (johangirod) [reply pages/definitions/[slug].tsx:45]: I'm not sure if there is a need for an additional abstraction for this line. Unless you meant something else ?

johangirod:
> Plus one thing that would really be lovely would be to add visual anchors like I did in https://api.gouv.fr/les-api/api-particulier#a-quoi-sert-l%E2%80%99api-particulier-?

Really simple to do so with this refacto, even if it means adding a new dependency : https://github.com/rehypejs/rehype-autolink-headings

XavierJp:
Is nt a dependency a bit overkilled for just a few line of code ?

johangirod:
Ok I choose to add a new dependency, `rehype-autolink-heading` for heading headlines : 
- It's the easiest and clearest way of doing it inside remark / rehype pipeline
- If I wanted to to without, and still use unified, I would have to install other utils dependencies anyway to transform the AST
- The other possibility is to move to react world and use `react-remark` to parse markdown and add transformation directly with react utilities (find headings in children, etc.)
  - The drawback, is the possibility to add this logic on the client side if we are not careful with our boundary, which would be a big overhead

I'm not fond of adding a lot of dependencies on a project either. But I think in that case, it's justified. If it becomes a hassle to upgrade, we can go back to `react-remark`

pages/definitions/[slug].tsx
Comment (XavierJp) [reply pages/definitions/[slug].tsx:37]: I meant the abstraction as I am not a big fan of `dangerouslySetInnerHTML={{ __html: definition.body.html }}` as it can easily break the whole page if the html is not valid

components-ui/summary/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+          <li>
+            <a className="fr-summary__link" href={`#${id}`}>
+              {content}{' '}
+            </a>
+          </li>
+        ))}
+      </ol>
+    </nav>
+  );
+}
```
XavierJp, line 32: Very nice

pages/lp/[slug].tsx
```diff
@@ -67,13 +67,21 @@ const LandingPage: NextPageWithLayout<IProps> = ({
           <div key={block.title} className="fr-col-12 fr-col-sm-4 fr-col-md-4">
             <div className="reassurance">
               <h2>{block.title}</h2>
-              <ReactMarkdown>{block.body}</ReactMarkdown>
+              <div
+                dangerouslySetInnerHTML={{
+                  __html: parseMarkdownSync(block.body).html,
+                }}
+              />{' '}
```
Comment (XavierJp) [inline pages/lp/[slug].tsx:74]: You repeat several time the same block which : 
- do tricky things (`dangerouslySetInnerHtml` is a bit dangerous)
- import `parseMarkdownSync`

I would feel more comfortable with something  like : 

`<RenderMarkdown  data={block.body} />`

Where `parseMarkdownSync` is called internaly and more documented and where data is typed using `IParsedMakdown`

components/markdown/RenderMarkdownServerOnly.tsx
```diff
@@ -0,0 +1,24 @@
...
+export default function RenderMarkdownServerOnly({
+  children,
+  showToc = false,
+}: {
+  showToc?: boolean;
+  children: string;
+}) {
+  if (typeof window !== 'undefined') {
+    throw new Error('This component is server-side only');
+  }
```
XavierJp, line 13: Nice safegard

components/markdown/RenderMarkdownServerOnly.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  return (
+    <>
+      {showToc && (
+        <Summary headings={headings.filter(({ depth }) => depth === 2)} />
+      )}
+      <div dangerouslySetInnerHTML={{ __html: html }} />
+    </>
+  );
+}
```
XavierJp, line 24: Simple but 👍

frontend/style/globals.css
```diff
@@ -9,6 +9,7 @@ a {
 
 body,
 html {
+  --text-spacing: 1.2rem 0;
   margin: 0;
   padding: 0;
```
XavierJp, line 14: Sounds like these few lines can have a huge impact. If you are confident with the change let's go with it

frontend/style/globals.css
```diff
@@ -146,3 +143,18 @@ p {
...
+/* # that appears before headings in content extracted from markdown */
+.anchor-link {
+  font-size: 1em;
+  float: left;
+  margin-left: -1em;
+  padding-right: .25em;
+  color: #000091;
+  opacity: 0;
+  background-image: none !important;
+}
```
XavierJp, line 156: I generally avoid `em` and always prefer `rem` but it does sound like this is one of the few case where `em` is more relevant 👍

pages/definitions/index.tsx
```diff
@@ -17,17 +14,19 @@ const AllDefinitionsPage: NextPageWithLayout<{
       canonical="https://annuaire-entreprises.data.gouv.fr/definitions"
     />
     <StructuredDataFAQ
-      data={definitions.map(({ title, body }) => [
-        title,
-        renderToStaticMarkup(<ReactMarkdown>{body}</ReactMarkdown>),
-      ])}
+      data={definitions.map(({ title, body }) => [title, body.html])}
```
XavierJp, line 17: We should use parseMarkdownSync here

But I dont know it it makes sense to still use `StructuredDataFAQ` as these pages looks increasingly like content pages and it is too long.

pages/faq/index.tsx
```diff
@@ -1,26 +1,20 @@
...
     <Meta
       title="FAQ de l'Annuaire des Entreprises"
       canonical="https://annuaire-entreprises.data.gouv.fr/faq"
     />
     <StructuredDataFAQ
-      data={articles.map(({ title, body }) => [
-        title,
-        renderToStaticMarkup(<ReactMarkdown>{body}</ReactMarkdown>),
-      ])}
+      data={articles.map(({ title, body }) => [title, body.html])}
```
XavierJp, line 17: We should use `parseMarkdownSync` here

pages/historique-des-modifications.tsx
```diff
@@ -26,17 +25,17 @@ const Changelog: NextPageWithLayout<IProps> = ({ changelog }) => (
...
-            </ReactMarkdown>
+            />
             {change.description && (
-              <i>
-                <ReactMarkdown>{change.description}</ReactMarkdown>
-              </i>
+              <em
+                dangerouslySetInnerHTML={{
+                  __html: parseMarkdownSync(change.description).html,
+                }}
```
XavierJp, line 37: Needs to be updated with the new logic

utils/static-pages/load-all.ts
```diff
@@ -0,0 +1,21 @@
...
+    // weirdly context add duplicates - this filter removes them
+    .filter((k: string) => k.indexOf('./') === 0)
+    .forEach((key: string, index: number) => {
+      const slug = key.replace('.yml', '').replace('./', '');
+      //@ts-ignore
+      articles.push({ ...values[index], slug });
+    });
+
+  return articles;
+}
```
XavierJp, line 21: I think these can be made even more generic and become a loader for all markdown collections but maybe this is a topic for another PR

XavierJp (approved):
I think it can be merged as soon as the build is fixed

frontend/style/globals.css
johangirod: I copied your code from data.gouv, but yeah, I agree that its better using `em` since we want to adapt to the container font-size.

frontend/style/globals.css
johangirod: I've browse a fair amount of pages and I didn't see any visual regression. This allows to remove custom `p` `margin` which was messing with the default style of `summary-fr` (TOC)

frontend/style/globals.css
XavierJp: hahaha I was wondering ! And I probably copied it myself from a smart programmer somewhrere 🤣

XavierJp:
Also : if you rebase with main, preview app should work

PR #726, opened 2023-10-25
chore: improve errors on sentry

johangirod:
- [x] refactor: use nextJS sentry in front when useReact
  - Add a common entry in front that is used in pages with react (which doesn't contain sentry browser init)
  - Remove frontend index.js load in page with react enabled
- [x] chore: use default nextJS config for sentry
  - use server.config.ts, client.config.ts and edge.config.ts like in the doc
  - add a withSentryConfig to next.config.js, which select the right config file depending on the env and add sourcemap to sentry
  - replace isSentryInitialized by isSentryActivated
  - the front & back error will now be sent to the sentry project : annuaire-entreprises-back (we might want to rename it to annuaire-entreprises-site)
- [x] [fix: log error instead of message in redis-storage](https://github.com/etalab/annuaire-entreprises-site/pull/726/commits/9d5ff39abfcb0e29603579c7b4d5cbc2fa3b0cd9) so that we have the stacktrace and explicit message in sentry
- [ ] ci: add new env variable to GitHub & production machine

You can review commit by commit for this PR.


frontend/src/load-bar.js
```diff
@@ -40,9 +40,6 @@ const init = () => {
 };
 
 const loadBarFactory = () => {
-  const _loader = null;
-  let _currentJobId = null;
```
XavierJp, line 44: Why delete these two var ?

frontend/src/common.js
```diff
@@ -0,0 +1,3 @@
+import './helpers-copy-paste';
+import './helpers-feedbacks';
+import './load-bar';
```
XavierJp, line 3: Great. 

I believe we also need front state machine and inpi pdf download

Can you add a detailled comment to explain the purpose of this file and distinguish it from `/index` ?

frontend/src/sentry.js
```diff
@@ -1,8 +1,10 @@
 import * as Sentry from '@sentry/browser';
 
-var dsn = import.meta.env.VITE_SENTRY_FRONT_DSN;
+const dsn = import.meta.env.VITE_SENTRY_FRONT_DSN;
+export const isSentryActivated = import.meta.env.PROD && Sentry && dsn
```
XavierJp, line 4: I fear there might be a possible confusion between the two `isSentryActivated` maybe we could have distinct names like `isSentryReactActivated`vs `isSentryViteActivated`

sentry.edge.config.ts
```diff
@@ -0,0 +1,9 @@
+import * as Sentry from '@sentry/nextjs';
+import { isSentryActivated } from '#utils/sentry';
+if (isSentryActivated) {
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
+    tracesSampleRate: 0.1,
+    maxBreadcrumbs: 0, // dont log breadcrumb
+  });
+}
```
XavierJp, line 9: Is it necessary when we dont have edge functions ?

frontend/src/load-bar.js
johangirod: They were listed as not used by typescript, so I removed them without thinking. And indeed, const & let are not assigned to `this`.

sentry.edge.config.ts
johangirod: I agree, it shouldn't be there. There was a warning without it, but I'm going to investigate further and see if we can do without.

frontend/src/common.js
johangirod: https://github.com/etalab/annuaire-entreprises-site/pull/726/commits/f6c8ff2c5057f6f16de9d16004aebd5913743559#diff-ff1d61de8655c32a447df582516565f722b8f17d0054fade55829de6800d9a2fR1-R10

sentry.edge.config.ts
johangirod: Well, I just removed it, the warning is just a warning that we can safely ignore while we're not using Edge functions

XavierJp:
Thanks I will test it, add env variable and merge it tomorrow

Review app are going to be a blast for this kind of PR

XavierJp (approved):

PR #743, opened 2023-10-30
fix: deactivate sentry on build

johangirod:


XavierJp (approved):
Wont it deactivate Sentry entirely ?

PR #746, opened 2023-11-01
fix: use canonical link in <Title /> section

XavierJp:
- always use full link for "Fiche résumé" tab

This should improve a bit on #730


johangirod (approved):
Bien vu !

PR #747, opened 2023-11-01
feat: move finances to the client

XavierJp:
Parial implementation of #686 focused on données-financières.

As this page triggers a lot of hydration errors I decided to migrate it ahead of the rest of ODC clients.

To read carefully :
- I stumbled upon the `stubbed` client as it doesnot work well with frontend and lazilly commented it ! (need a better solution)
- I added support for `params` in frontend client which might be useful in the future.


johangirod (commented):
This is a definite limit of `stubClient` : it cannot stub client on front-end yet. I'll have to improve it to handle this case.

johangirod (approved):

PR #752, opened 2023-11-02
refactor: move BrowserIsOutdatedBanner from layout

johangirod:
This PR refactor the BrowserIsOutdatedBanner component from the layouts component into the generic one in _app.tsx

It also changes the message to present what is working instead of what is not.

### TODO 

- investigate why the warning message is displayed for IE 11 and not for firefox < 65 in prod only (it works in local and in preview apps)

### Screenshot

image

### DIFF
--- a/components/banner/browser-is-outdated.tsx
+++ b/components/banner/browser-is-outdated.tsx
@@ -10,20 +10,26 @@ export const BrowserIsOutdatedBanner: React.FC<{}> = () => (
       aria-label="Votre navigateur est obsolète"
     >
       <div className="fr-container">
-        <b>⚠️ Votre navigateur est obsolète : </b>
-        le site est utilisable mais certaines fonctionnalités comme le
-        déclenchement automatique des téléchargements, la recherche avancée ou
-        le copier/coller ne fonctionneront pas.
-        <br />
-        Avoir un navigateur à jour est{' '}
-        <a
-          target="_blank"
-          rel="noreferrer noopener"
-          href="https://www.ssi.gouv.fr/entreprise/precautions-elementaires/bonnes-pratiques-de-navigation-sur-linternet/"
-        >
-          recommandé par l’ANSSI pour naviguer sur internet en sécurité
-        </a>
-        .
+        <h1>⚠️ Votre navigateur est obsolète </h1>
+        <p>
+          La plupart des fonctionalités de ce site ne fonctionneront pas. Vous
+          pourrez uniquement :
+        </p>
+        <ul>
+          <li>Rechercher une entreprise par son nom ou son SIREN</li>
+          <li>Consulter la fiche résumé de l’entreprise</li>
+        </ul>
+        <p>
+          Avoir un navigateur à jour est{' '}
+          <a
+            target="_blank"
+            rel="noreferrer noopener"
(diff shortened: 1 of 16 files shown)


XavierJp:
Really like what you did with the refacto. I wonder if we can even move it out of `postServerSideProps` ?

XavierJp:
Interesting read, but do not seem to take very old browser into account : 
https://andreasremdt.com/blog/detecting-outdated-browsers/

Another idée : use feature detection and triger banner from frontend : 
https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection

Possible links (but private companies) : 
https://browser-update.org/update-browser.html
https://bestvpn.org/outdatedbrowser/en.html

components/banner/browser-is-outdated.tsx
```diff
@@ -32,12 +32,29 @@ export const BrowserIsOutdatedBanner: React.FC<{}> = () => (
...
+      //  https://caniuse.com/?search=esmodule
+      //  https://stackoverflow.com/questions/44891421/detect-es6-import-compatibility
+      dangerouslySetInnerHTML={{
+        __html: `
+          const esModuleDisabled = typeof window !== 'undefined' 
+            && !('noModule' in HTMLScriptElement.prototype)
+            
+          if (esModuleDisabled) {
+            getElementById('browser-is-outdated').style.display = 'block'
+          }
```
XavierJp, line 47: Not sure `const`, `&&` and `style.display` is backward compatible with very old browsers. At least we should encapsulate iin a try catch that would default on displaying the banner

package.json
```diff
@@ -39,7 +39,6 @@
     "@types/qrcode": "^1.5.2",
     "@types/react": "^18.2.28",
     "@types/react-dom": "^18.2.13",
-    "@types/ua-parser-js": "^0.7.37",
```
XavierJp, line 42: I love when we remove packages <3

XavierJp (changes requested):
I fear this might not work on very old browsers

components/banner/browser-is-outdated.tsx
johangirod: I think we're safe for && and style
- https://caniuse.com/mdn-html_elements_style
- https://caniuse.com/mdn-javascript_operators_logical_and

I can replace const for var to allow IE < 11

components/banner/browser-is-outdated.tsx
```diff
@@ -5,38 +5,65 @@ import constants from '#models/constants';
...
+      //  https://caniuse.com/?search=esmodule
+      //  https://stackoverflow.com/questions/44891421/detect-es6-import-compatibility
+      dangerouslySetInnerHTML={{
+        __html: `
+          var esModuleDisabled = typeof window !== 'undefined' 
+            && !('noModule' in HTMLScriptElement.prototype)
+            
+          if (esModuleDisabled) {
+            getElementById('browser-is-outdated').style.display = 'block'
+          }
```
XavierJp, line 47:
```suggestion
	  try {
             var esModuleDisabled = typeof window !== 'undefined' 
               && !('noModule' in HTMLScriptElement.prototype);
            
             if (esModuleDisabled) {
                 throw new Error('browser is outdated');
          } catch {
             getElementById('browser-is-outdated').style.display = 'block';
          }
```

XavierJp (approved):

PR #755, opened 2023-11-06
fix: add og:type for twitter card generation

johangirod:


johangirod:
Can't check on review app because of robot.txt but I think it should do it :crossed_fingers:

XavierJp (approved):

PR #756, opened 2023-11-06
Better data section loading UX

johangirod:
- [feat: add a message when data loading takes long](https://github.com/etalab/annuaire-entreprises-site/commit/9d83fbf2a606ff8c097df299b095da7648fe584d)
- [feat: change insee and rne sections order](https://github.com/etalab/annuaire-entreprises-site/commit/1cb23ee617f023f6ab39cee11c6e4ba1088f241b)
- [fix: remove height transition on data loader](https://github.com/etalab/annuaire-entreprises-site/commit/ecb34271dca2cfcf5f55baf7ecb8405ebbcd5a35)

### DIFF
--- a/components-ui/horizontal-separator/index.tsx
+++ b/components-ui/horizontal-separator/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import BreakPageForPrint from '#components-ui/print-break-page';
 import constants from '#models/constants';
 
 export const SimpleSeparator = () => (
@@ -17,6 +17,7 @@ export const SimpleSeparator = () => (
 
 export const HorizontalSeparator = () => (
   <>
+    <BreakPageForPrint />
     <div className="horizontal-separator layout-center">
       <span className="line" />
       <span className="circle" />
--- a/components-ui/horizontal-separator/index.tsx
+++ b/components-ui/horizontal-separator/index.tsx
@@ -1,4 +1,3 @@
-import BreakPageForPrint from '#components-ui/print-break-page';
 import constants from '#models/constants';
 
 export const SimpleSeparator = () => (
@@ -17,7 +16,6 @@ export const SimpleSeparator = () => (
 
 export const HorizontalSeparator = () => (
   <>
-    <BreakPageForPrint />
     <div className="horizontal-separator layout-center">
       <span className="line" />
       <span className="circle" />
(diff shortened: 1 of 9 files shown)


components-ui/horizontal-separator/index.tsx
```diff
@@ -17,6 +17,7 @@ export const SimpleSeparator = () => (
 
 export const HorizontalSeparator = () => (
   <>
+    <BreakPageForPrint />
     <div className="horizontal-separator layout-center">
       <span className="line" />
       <span className="circle" />
```
XavierJp, line 23: This might create empty blank page where we called `<HorizontalSeparator/>`next to `<BreakPageForPrint />`
Also, we sometimes use several separator pretty closelys.

But I like the idea. If we dfeploy this, I would use a `<PrintNever />` on the separator itself as it is no longer required

components/immatriculations/index.tsx
```diff
@@ -51,6 +52,9 @@ const Immatriculations: React.FC<IProps> = ({
         immatriculationRNE={immatriculationRNE}
         uniteLegale={uniteLegale}
       />
+      <AvisSituationSection uniteLegale={uniteLegale} session={session} />
+      <HorizontalSeparator />
+
```
XavierJp, line 57: Why move this up ? 

On a business point of view it does not really make sense as the association immatriculation is more equivalent to `immatriculationRNE` and therefore is expected to come first

Plus I expect most people would open thise page for the `immatriculation` rather than the `AvisSituation`

johangirod: I explained it in the commit : 
```
In `Dirigeants` section, we display the `Avis de situation` section
before the `RNE` section. The reason is that the inpi often takes very
long to load. In order to improve UX, we display the `Avis de situation`
which is already loaded. This means that the UI changes happens outside
of the viewport.
```

But I understand the point about people wanting to see the immatriculation RNE first. Besides, it changes the way things are presented which can be annoying for people that are used to it. I let you decide if you think it's better to skip this commit I'm totally OK.

components/section/data-section-loader.tsx
```diff
@@ -28,6 +33,14 @@ export default function DataSectionLoader({
...
-    </div>
+      {after5s && (
+        <FadeIn>
+          <Warning>
+            Le service semble particulièrement occupé en ce moment, cela peut
+            prendre encore un peu de temps (10s à 20s).
+          </Warning>
+        </FadeIn>
+      )}
+    </>
```
XavierJp, line 44: For UX coherence I would add it before as it is the way it is displayed accross the app

XavierJp (approved):
I revert the changes on `<HorizontalSeparator />`
I think we should put the warning at the begining of the section

PR #760, opened 2023-11-06
Move JOAFE calls to client

johangirod:
- add useFetch hook to factorize fetching data logic
- move JOAFE fetch to client side
- remove stubs on JOAFE and DCA clients (TODO)

references #686


components/annonces-section/comptes-association.tsx
```diff
@@ -7,19 +7,24 @@ import { DataSection } from '#components/section/data-section';
...
+    | IComptesAssociation
+    | IAPINotRespondingError
+    | IAPILoading;
 }> = ({ association, comptesAssociation }) => {
   return (
     <DataSection
       data={comptesAssociation}
       title="Dépôt de Comptes des Associations"
       sources={[EAdministration.DILA]}
+      notFoundInfo={null}
```
XavierJp, line 27: notFoundInfo can be `undefined` why specify `null` here and not just ignore it ?

components/annonces-section/annonces-association.tsx
```diff
@@ -8,19 +8,24 @@ import { DataSection } from '#components/section/data-section';
...
+    | IAnnoncesAssociation
+    | IAPINotRespondingError
+    | IAPILoading;
 }> = ({ association, annoncesAssociation }) => {
   return (
     <DataSection
       title="Annonces Journal Officiel des Associations"
       sources={[EAdministration.DILA]}
       data={annoncesAssociation}
+      notFoundInfo={null}
```
XavierJp, line 28: notFoundInfo can be `undefined` why specify `null` here and not just ignore it ?

pages/annonces/[slug].tsx
```diff
@@ -5,36 +5,32 @@ import { ComptesAssociationSection } from '#components/annonces-section/comptes-
...
 }
 
 const Annonces: NextPageWithLayout<IProps> = ({
-  annoncesAssociation,
-  comptesAssociation,
   metadata: { session },
   uniteLegale,
 }) => {
+  const annoncesAssociation = useFetchJOAFE(uniteLegale);
+  const comptesAssociation = useFetchComptesAssociation(uniteLegale);
```
Comment (XavierJp) [inline pages/annonces/[slug].tsx:33]: Why not have these two call into the `AnnoncesAssociationSection` and `ComptesAssociationSection` ?

XavierJp (approved):
Sounds good, just a bit suprised by the fact you call the hooks on parents component

components/annonces-section/comptes-association.tsx
johangirod: If `null` it means that the block will disappear if no data is found, which was the behavior here before we moved the call to front. 

If not specified (`undefined`), the `Section` is displayed with a generic error message (`AdministrationNotRespondingMessage`).

What's at stakes is to choose if the behaviour « not found » is abnormal (an error is displayed), or if it can happen and just means the association is not concerned (we hide the block).

That being said, I'm not 100% fan of the « hide the block » behaviour, because it looks like a glitch. Something is loading and then suddenly, it's not. It'll be better to let the `Section` block and add a message like « l'association n'a pas déposé de comptes ». 

In short, I'm in favour of removing the `null` behaviour, and force display of a message anyway.

PR #762, opened 2023-11-07
chore: add annuaire-entreprises as user agents for our providers

XavierJp:
- fix anchors style in md
- update supported browser as we use `?.` a LOT
- better user agent for outgoing calls


johangirod (approved):
Good catch for the optional chaining, which is more restrictive than esmodule

PR #778, opened 2023-11-13
chore: improve sentry

johangirod:
- Remove sensitive headers
- Add is_bot tag


XavierJp (approved):
Nice. No need for a try catch ?

johangirod:
Just in case, I added one around `isUserAgentABot`  :+1:

PR #787, opened 2023-11-16
feat: migration to updownio API

### LINKED ISSUE #773: Renouvellement de la facture UptimeRobot le 18 Novembre
- [ ] Test de updown io

XavierJp:
This is a migration proposal from Uptime Robot to Updown.io

### The good : 
- free (shared account for beta/dinum)
- simple
- we can use tchap webhooks and even telegram for "astreinte"

### The bad :
- shared account so we cannot manage it and so far the process to add modify a monitor is quite complicated 
- require an API call per monitoring and we still dont have as much data as uptime robot (we miss the creation date, that would require an aditional call)
- not sure we can organise monitoring page the way we want (see https://updown.io/p/3sg8p) 

closes #773


XavierJp:
@BenoitSerrano could not find the API Bilans Financiers in the list of monitors. 

I strongly believe I need an access to updown io or this migration will become a nightmare

components/api-monitoring/index.tsx
```diff
@@ -22,9 +19,6 @@ const getUptimeColor = (ratio: IRatio) => {
 };
 
 const getUptimeLabel = (ratio: IRatio) => {
-  if (!ratio.isActive) {
-    return 'Aucune donnée sur cette période.';
-  }
```
XavierJp, line 27: So for instance we loose the concept of no data because the monitor did not exist yet.

models/monitoring.ts
```diff
@@ -25,43 +24,49 @@ export interface IMonitoringWithMetaData extends IMonitoring {
...
+                  apiSlug,
+                  apiName,
+                  administrationEnum,
+                  ...monitoring,
+                } as IMonitoringWithMetaData;
+              } catch (e: any) {
+                logErrorInSentry(e, {
+                  errorName: 'Error while fetching monitoring',
+                });
+                throw e;
```
XavierJp, line 62: Not sure about all this @johangirod : 

- not sure this is an actual promise all and not sequential calls
- not sure about the behaviour when it fails. Should we wrap it into `APINotResponding` ? As the page is cached and revalidated every 30 seconds, I believe it is better to let it fail. But as we do a LOT of call to their API I wonder if this is a good idea

models/monitoring.ts
```diff
@@ -1,20 +1,19 @@
-import { clientMonitorings } from '#clients/monitoring';
+import { clientMonitoring } from '#clients/monitoring';
 import logErrorInSentry from '#utils/sentry';
 import { administrationsMetaData } from './administrations';
+import { IAPINotRespondingError } from './api-not-responding';
 
 export type IRatio = {
+  ratioNumber: number;
   ratio: string;
```
johangirod, line 8: Not sure the `ratio` should have both `ratio` and `ratioNumber`. I'm in favour of removing the `string` property and provide only the `number` one. It's the responsability of the view to choose how to display the data (how many decimals for instance).

models/monitoring.ts
johangirod: `await` works inside `for` loop but it will end up in sequential call, waiting for the current iteration to resolve before starting the next one. I think it would be clearer to construct the returned object doing the following : 

1. Get all API monitor ids (monitorIds) 
2. Construct an object `Array<Promise<id, IMonitor | IAPINotResponding >`
3. `Object.fromEntries(Promise.all()) de cet objet
4. Reconstruire l'objet de sortie à partir de cela

johangirod (approved):

components/api-monitoring/index.tsx
```diff
@@ -22,9 +19,6 @@ const getUptimeColor = (ratio: IRatio) => {
...
   const uptimeNum = parseFloat(ratio.ratio);
```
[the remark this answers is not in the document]
XavierJp: Oh my bad I see

XavierJp:
@johangirod I took all of your comments into account. Thanks for the explanation about the `for` loop vs `Promise.all`. Makes much more sesn this way. I am ready to deploy.

PR #792, opened 2023-11-20
Improve error recovering and logging

johangirod:
### [feat: add error boundary for client errors](https://github.com/etalab/annuaire-entreprises-site/commit/05f29e5130703d5b2a0246014e057686f661ddd5)

This commit adds an error boundary to catch client errors and display a
message to the user.

It allows to pass additionnal metadata to Sentry (like the component
stack), and is the [recommanded
way](https://nextjs.org/docs/pages/building-your-application/configuring/error-handling)
to handle clients errors in Next.js.

Besides, it changes the error message for server errors to be less scary
(remove the "⚠️").

### [fix: safari 13 error](https://github.com/etalab/annuaire-entreprises-site/commit/14123547ac5883bec4be46d6c1564f85c2e2ec3b)

fix https://errors.data.gouv.fr/organizations/sentry/issues/129195

### [fix: add public class to required syntax](https://github.com/etalab/annuaire-entreprises-site/commit/1547c33c217bc9fda3019ab2ab56613143708631)

Cf https://caniuse.com/?search=mdn-javascript_classes_public_class_fields

### [feat: add error boundary to data sections](https://github.com/etalab/annuaire-entreprises-site/commit/65bd8f3f93b6c97a2c1d64961190492a6819df4f)

It enable to display an error message when a section throw an error only
in the section component instead of crashing the whole app.


components/banner/browser-is-outdated.tsx
```diff
@@ -34,14 +34,18 @@ export const BrowserIsOutdatedBanner: React.FC<{}> = () => (
...
     <script
-      //  Show warning for browsers without Optional chaining
-      //  https://caniuse.com/mdn-javascript_operators_optional_chaining
-      //  https://stackoverflow.com/questions/66288922/check-if-optional-chaining-is-supported
+      //  Show warning for browsers
+      //  1. without Optional chaining
+      //    - https://caniuse.com/mdn-javascript_operators_optional_chaining
+      //    - https://stackoverflow.com/questions/66288922/check-if-optional-chaining-is-supported
+      //  2. without public class fields
+      //    - https://caniuse.com/?search=mdn-javascript_classes_public_class_fields
```
XavierJp, line 42: Very interesting, the browser list difference is very subtle

components/section/section-error-boundary.tsx
```diff
@@ -0,0 +1,51 @@
...
+                <a
+                  target="_blank"
+                  rel="noreferrer noopener"
+                  href="https://www.ssi.gouv.fr/entreprise/precautions-elementaires/bonnes-pratiques-de-navigation-sur-linternet/"
+                >
+                  fortement recommandé par l’ANSSI
+                </a>{' '}
+                pour naviguer sur internet en sécurité.
+              </p>
+            </>
```
XavierJp, line 30: I think this is overkill as if `window.IS_OUTDATED_BROWSER` is `true` then user should already have this message on top of the page

pages/_app.tsx
```diff
@@ -23,9 +24,9 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
   //eslint-disable-next-line react/jsx-props-no-spreading
   const layout = getLayout(<Component {...pageProps} />, session);
   return (
-    <>
+    <ErrorBoundary>
       <BrowserIsOutdatedBanner />
       {layout}
-    </>
+    </ErrorBoundary>
```
XavierJp, line 30: I fear this might break stuff, especially the redirect on error logic

johangirod: Only on client. And that's the goal, we want to differentiate between 500 errors (serverside) and client error :
- On client, add other metadata (like component call stack)
- It should fix the reporting of error, which is broken on the client 

See: 
- [Error that are not properly catched](https://errors.data.gouv.fr/organizations/sentry/issues/129277/?project=15&query=is%3Aunresolved+assigned%3Ame&referrer=issue-stream&statsPeriod=14d)
- [What next does say about them](https://nextjs.org/docs/messages/client-side-exception-occurred)
- [What is the recommanded way in the doc](https://nextjs.org/docs/pages/building-your-application/configuring/error-handling)
- « Error boundaries do not catch errors for Server side rendering » in [react documentation](https://legacy.reactjs.org/docs/error-boundaries.html)

frontend/src/sentry.js
```diff
@@ -32,7 +31,7 @@ if (isViteSentryActivated) {
         },
       }),
     ],
-
+    tracesSampleRate: 0.005,
```
XavierJp, line 34: 👍

XavierJp (approved):

PR #804, opened 2023-11-28
test: update snapshots & add automatic PR 

johangirod:
- Update E2E & snapshot test following Manakin Production is no longer entrepreneur du spectacle
- Add automatic PR for api client test


johangirod:
> Sad to see Manakin Prod loosing it s licence :)

It's an end of an era :p

PR #815, opened 2023-12-08
feat: stop using insee for UL with more than 100 etablissements

XavierJp:
Congrats Hajar for the amazing work


johangirod:
The API client test would have to be updated, but I can do it.

PR #820, opened 2023-12-11
fix: add original error message if none present

johangirod:


XavierJp (approved):

XavierJp:
@johangirod added another fix

hooks/fetch/finances-societe.ts
```diff
@@ -12,19 +12,17 @@ export function useFetchFinancesSociete(uniteLegale: IUniteLegale) {
...
-          context: {
-            siren: uniteLegale.siren,
-          },
-        });
-        if (e.status === 404) {
-          logInfoInSentry(exception);
-        } else {
-          logErrorInSentry(exception);
-        }
+        if (e.status !== 404)
```
johangirod, line 15: Yep, i hesitated to log 404. I though, it could be interresting to follow if we do not have an unusual number of errors due to a bug in their APIs. But in the same time it's a lot of noise. There is also bodacc, joaffe and finance-société which logs 404.

XavierJp: As we have a "call all the time just in case" poolicy, we will have a LOT of 404 for these 3 API. Ithink we should make 404 silent for BODACC and JOAFE also

PR #832, opened 2023-12-15
fix: fetch resource exception is not defined

XavierJp:
Unsure what to do here when `FetchRessourceException is not defined`

https://errors.data.gouv.fr/organizations/sentry/issues/131620/?project=16&query=is%3Aunresolved&referrer=issue-stream

I think it is dangerous to share code between vite and react. I would advise using a custom exception here. 

I would be more confortable if you handle it from here @johangirod

### DIFF
--- a/frontend/src/fetch-tva.js
+++ b/frontend/src/fetch-tva.js
@@ -33,12 +33,7 @@ import { extractSirenSlugFromUrl, formatIntFr } from './utils';
         // We dont log errors, as they are already logged in the backend
         tvaContainer.setError();
         if (e instanceof TypeError) {
-          throw new new FetchRessourceException({
-            ressource: 'VerifyTVA',
-            cause: e,
-            context: { siren },
-            administration: EAdministration.VIES,
-          })();
+          throw e;
         }
       });
   }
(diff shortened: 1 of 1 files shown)


frontend/src/fetch-tva.js
```diff
@@ -33,12 +33,7 @@ import { extractSirenSlugFromUrl, formatIntFr } from './utils';
         // We dont log errors, as they are already logged in the backend
         tvaContainer.setError();
         if (e instanceof TypeError) {
-          throw new new FetchRessourceException({
-            ressource: 'VerifyTVA',
-            cause: e,
-            context: { siren },
-            administration: EAdministration.VIES,
-          })();
+          throw e;
```
johangirod, line 36: I would rather instantiate a new error to prevent the multiple duplicates in sentry due to browser specificity.

```suggestion
          throw new Error("Client error while fetching TVA", {cause : e});
```

johangirod (approved):
Oops, my bad. It's a good thing indeed to keep the codebase separated. We might want to add typescript checking on JS files in order to catch those mistakes before production (if we modify the `frontend/` directory again).

frontend/src/fetch-tva.js
XavierJp: Seems good to me

johangirod (approved):

PR #838, opened 2023-12-18
Add a form to hide personal data on company page

johangirod:
Fix #676 

## Todo 
- [x] Change FAQ
- [x] Add env variables to prod & scalingo
- [x] Recette FranceConnect
- [x] Change URL
- [x] Integrate FranceConnect
- [x] Create a table in Grist
- [x] Manualluy test all cases
- [x] Add a logout functionality

PR #848, opened 2024-01-04
feat: create agent connect oidc client

### LINKED ISSUE #816: [Agent] Migrer MCP vers AgentConnect
- Migrer MCP vers AgentConnect
- Récupérer les siret des organisations et envoyer le siret à API Entreprise

### LINKED ISSUE #852: [Agent Connect] connect/disconnect doit retomber sur la page

XavierJp:
closes #816 
closes #852


clients/auth/agent-connect/strategy.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    return _client;
+  }
+};
+
+export const agentConnectAuthorizeUrl = async (req: IncomingMessage) => {
+  const client = await getClient();
+
+  const nonce = generators.nonce();
+  const state = generators.state();
```
johangirod, line 48: Nice, didn't know about those utils, I'm gonna use them in #838

hooks/use-siren-from-url.ts
```diff
@@ -0,0 +1,11 @@
+import { useRouter } from 'next/router';
+import { extractSirenOrSiretSlugFromUrl } from '#utils/helpers';
+
+export default function useSirenFromRouter(): string | null {
```
johangirod, line 4: Why use `sirenFrom` and not `pathFrom`? It seems to me that going redirecting to the exact page where the user started the auth workflow provides a better experience…

utils/session/index.ts
```diff
@@ -47,6 +53,27 @@ export const setAgentSession = async (
...
+
+export const setSirenFrom = async (session: IronSession, sirenFrom: string) => {
+  session.sirenFrom = sirenFrom;
+  await session.save();
+};
+
+export const getSirenFrom = (session: IronSession) => session.sirenFrom;
+
+export const cleanSirenFrom = async (session: IronSession) => {
+  session.sirenFrom = undefined;
```
johangirod, line 74:
```suggestion
  delete session.sirenFrom;
```
I find it more bulletproof, otherwise `sirenFrom in session` would return `true`. But it's not crucial, and if your code typecheck, it must mean that I'm missing something and that I may be mistaken…

johangirod (approved):

hooks/use-siren-from-url.ts
XavierJp: I agree it is more powerful but it can becomes complicated for logout as the pathFrom is then forbidden 

I suggest we do this in a dedicated PR

PR #854, opened 2024-01-09
feat: add a loginWall on documents sections

### LINKED ISSUE #818: [Agent] Premier déploiement du bouton espace agent
Ajouter un CTA agents "contextuel" :
- [ ] alerte non diffusible en haut de page pour les agents 
- [x] page "Documents" avec option vers l'Inpi pour les non agents. Une <DataSection /> floutée pourrait être une bonne solution ici 
- [x] page comptes pour les documents des comptes et pour les données en confidentialité partielle
- [ ] section justificatifs pour les siren protégés et les non diffusibles

XavierJp:
closes #818


pages/documents/[slug].tsx
```diff
@@ -46,22 +47,20 @@ const UniteLegaleForAgentPage: NextPageWithLayout<IProps> = ({
...
       )}
-      {isAgent(session) && <ActesSection uniteLegale={uniteLegale} />}
+      <DocumentWall
+        isVisible={isAgent(session)}
+        title="Actes et statuts"
+        id="actes"
+        uniteLegale={uniteLegale}
+      >
+        <DocumentActesSection uniteLegale={uniteLegale} />
+      </DocumentWall>
```
Comment (XavierJp) [inline pages/documents/[slug].tsx:57]: @johangirod would like your input here. not a big fan of current implementation. How would you do that ?

components/espace-agent-components/documents/document-wall.tsx
```diff
@@ -0,0 +1,106 @@
...
+const DocumentWall: React.FC<
+  PropsWithChildren<{
+    isVisible: boolean;
+    title: string;
+    id: string;
+    uniteLegale: IUniteLegale;
+  }>
+> = ({ uniteLegale, id, title, isVisible = false, children }) => {
+  if (!isVisible) {
+    return (
```
johangirod, line 18: It's considered good practice to return early with the simplest case, to avoid unnecessary indentation. 
```suggestion
  if (isVisible) {
    return children
  }
  return (
```

pages/documents/[slug].tsx
Comment (johangirod) [reply pages/documents/[slug].tsx:57]: I think it's okay. The other way would be to add this logic to the `Section` component to factorize title, administration and such. I think we can postpone the decision until next time we need that logic.

components/espace-agent-components/documents/document-wall.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Section title={title} id={id} sources={[EAdministration.INPI]}>
```
johangirod, line 19: I would keep the pink border for coherence. 
```suggestion
      <Section isProtected title={title} id={id} sources={[EAdministration.INPI]}>
```

components/espace-agent-components/documents/document-wall.tsx
```diff
@@ -0,0 +1,106 @@
...
+                Les <b>particuliers, salariés</b> et <b>entrepreneurs</b>,
+                peuvent consulter cette donnée sur l’onglet document de{' '}
+                <a href={routes.rne.portail.entreprise + uniteLegale.siren}>
+                  la page data.inpi.fr de cette entreprise
+                </a>
+                .
+              </div>
+              <br />
+            </div>
+          </div>
```
johangirod, line 40:
```suggestion
            <div>
              <h3>
                Vous êtes agent public ?
              </h3>
              <div className="layout-center">
                <ButtonMonComptePro />
              </div>
              <h3>Comment accéder à ces documents autrement ?</h3>
              <div>
                Les <b>particuliers, salariés</b> et <b>entrepreneurs</b>,
                peuvent consulter cette donnée sur l’onglet document de{' '}
                <a href={routes.rne.portail.entreprise + uniteLegale.siren}>
                  la page data.inpi.fr de cette entreprise
                </a>
                .
              </div>
              <br />
            </div>
          </div>
```

components/espace-agent-components/documents/document-wall.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <div className="blur" tab-index="-1">
```
johangirod, line 41:
```suggestion
          <div className="blur" tab-index="-1" aria-hidden>
```

components/espace-agent-components/documents/document-wall.tsx
```diff
@@ -0,0 +1,106 @@
...
+            top: 0;
+            z-index: 10;
+            height: 100%;
+            width: 100%;
+          }
+          .cta-actes > div {
+            background-color: #fff;
+            border-radius: 3px;
+            padding: 0 30px;
+            box-shadow: 0 0 20px 0 rgba(0.5, 0.5, 0.5, 0.1);
```
johangirod, line 82:
```suggestion
            box-shadow: 0 0 20px 0 rgba(0.5, 0.5, 0.5, 0.1);
            border: 1px solid rgba(0.5, 0.5, 0.5, 0.1);
```

johangirod (approved):

XavierJp:
@johangirod 

- Changed a bit implementation into, I believe, I simpler one
- Use protected Section and indeed, much better
- Tried <h3/> suggestion but I think it does not work (I find font-size changes does not bring good content hierarchy)

components/espace-agent-components/agent-wall/documents.tsx
```diff
@@ -0,0 +1,31 @@
...
+  id: string;
+  uniteLegale: IUniteLegale;
+}> = ({ uniteLegale, id, title }) => {
+  return (
+    <AgentWall
+      id={id}
+      title={title}
+      modalFooter={
+        <>
+          <b>Comment accéder à ces documents autrement ?</b>
```
johangirod, line 16: > I find font-size changes does not bring good content hierarchy

It's not really about `font-size` change, but more about using the proper tag that conveys the best semantic meaning to the text. In this case, for instance, this is clearly a title of the following section, and thus, we should use a `h<x>` tag. This is good practice for SEO, navigation, and accessibility.

You can, of course, change the appearance of the `hx` in this case if you like. Personally, I like the ability to decouple a tag from appearance that `styled-components` allows with the `as` prop. 

Lastly, the `b` tag is deprecated as it does not convey semantic meaning. It's better to use `strong`.

Source: https://developer.mozilla.org/fr/docs/web/HTML/Element/strong#b_ou_strong

johangirod (approved):

components/espace-agent-components/agent-wall/documents.tsx
XavierJp: Ok makes more sense. I also understand better your implem of #838 

I tend to focus more on UX, maybe too much. Especially as these pages are not indexed by robot. But they could be visited

XavierJp: I am going to add simple `<hn/>` and remove `<strong />`

components-ui/button-agent-connect/index.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
-import useSirenFromRouter from 'hooks/use-siren-from-url';
+import usePathFromRouter from 'hooks/use-path-from-router';
 
 const ButtonAgentConnect: React.FC<{}> = () => {
-  const sirenFrom = useSirenFromRouter();
+  const pathFrom = usePathFromRouter();
```
johangirod, line 5: :purple_heart:

XavierJp: Hehe seul problème : je me suis planté dans l'implem.

johangirod: Des tests E2E à l'occasion pour tester le parcours de connexion, ça pourrait être bien (comme c'est un parcours critique pour les utilisateurs).

XavierJp: Double plus 1

Helas, pour le moment agentConnect ne supporte pas le localhost

johangirod: On peut mocker AgentConnect, c'est peut-être plus safe vu les petits soucis de fiabilité du service en ce moment :upside_down_face:

PR #855, opened 2024-01-09
feat: add custom dimension for tracking usager type

### LINKED ISSUE #817: [Agent] Mesure des KPIs essentiels
Faire remonter dans les events (et les stats ?) les KPIs des agents authentifiés
- nombre de connections
- NPS

johangirod:
This commit adds a custom dimension to the Matomo tracking script to track the type of user that is visiting the website.

It set the custom dimension to "Agent connecté" or "Super-agent connecté" if the user is logged in as an agent or super-agent.

Otherwise, it sets the custom dimension to the value of the `radio-set-visitor-type` form field in the NPS form.

closes #817

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -1,4 +1,5 @@
 import Document, { Head, Html, Main, NextScript } from 'next/document';
+import { isAgent, isSuperAgent } from '#utils/session';
 
 const manifest = (
   process.env.NODE_ENV === 'production'
@@ -127,7 +128,8 @@ class CustomDocument extends Document {
     const isProd = process.env.NODE_ENV === 'production';
     const useReact =
       this.props['__NEXT_DATA__']?.props?.pageProps?.metadata?.useReact;
-
+    const session =
+      this.props['__NEXT_DATA__']?.props?.pageProps?.metadata?.session;
     const HeadToUse = useReact ? Head : CustomHead;
     return (
       <Html lang="fr">
@@ -141,6 +143,13 @@ class CustomDocument extends Document {
               dangerouslySetInnerHTML={{
                 __html: `
               var _paq = window._paq || [];
+              ${
+                isSuperAgent(session)
+                  ? `_paq.push(['setCustomDimension', '1', 'Super-agent connecté']);`
+                  : isAgent(session)
+                  ? `_paq.push(['setCustomDimension', '1', 'Agent connecté']);`
+                  : ''
+              }
               _paq.push(['trackPageView']);
               _paq.push(['enableLinkTracking']);
               (function () {
--- a/pages/formulaire/merci.tsx
+++ b/pages/formulaire/merci.tsx
@@ -1,9 +1,11 @@
+import { GetServerSideProps } from 'next';
 import { ReactElement } from 'react';
 import ButtonLink from '#components-ui/button';
 import { LayoutDefault } from '#components/layouts/layout-default';
+import { postServerSideProps } from '#utils/server-side-props-helper/post-server-side-props';
 import { NextPageWithLayout } from 'pages/_app';
 
-const ThanksPage: NextPageWithLayout = () => {
+const ThanksPage: NextPageWithLayout<{ usager: string }> = ({ usager }) => {
   return (
     <div id="layout">
       <main>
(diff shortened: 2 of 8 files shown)


pages/_document.tsx
```diff
@@ -127,7 +128,8 @@ class CustomDocument extends Document {
     const isProd = process.env.NODE_ENV === 'production';
     const useReact =
       this.props['__NEXT_DATA__']?.props?.pageProps?.metadata?.useReact;
-
+    const session =
+      this.props['__NEXT_DATA__']?.props?.pageProps?.metadata?.session;
```
XavierJp, line 132: I see an issue here : 
props get serialized by next but session is supposed to be crypted is nt it a big security breach to serialize the session into the props ? 

Could we maybe use an ENUM for the different agent levels or use another way to inform the `_document` that the session is agent/super-agent without serializing the entire session object ?

pages/_document.tsx
```diff
@@ -141,6 +143,13 @@ class CustomDocument extends Document {
               dangerouslySetInnerHTML={{
                 __html: `
               var _paq = window._paq || [];
+              ${
+                isSuperAgent(session)
+                  ? `_paq.push(['setCustomDimension', '1', 'Super-agent connecté']);`
+                  : isAgent(session)
+                  ? `_paq.push(['setCustomDimension', '1', 'Agent connecté']);`
+                  : ''
+              }
```
XavierJp, line 152: Especially as it only get used to get the dimension. Mayb in metadata have a field `matomoDimension` set and directly use it here

```suggestion
		const matomoDimension = this.props['__NEXT_DATA__']?.props?.pageProps?.metadata?.matomoDimension || null;
		
		...

              `_paq.push(['setCustomDimension', '1', ${matomoDimension}]);`
```

pages/formulaire/merci.tsx
```diff
@@ -15,6 +17,13 @@ const ThanksPage: NextPageWithLayout = () => {
           <ButtonLink to="/">Retourner au moteur de recherche</ButtonLink>
         </div>
       </main>
+      <script
+        dangerouslySetInnerHTML={{
+          __html: `
+          var _paq = window._paq || [];
+          _paq.push(['setCustomDimension', '1', "${usager}"]);`,
+        }}
+      />
```
XavierJp, line 26: I dont see the need for this. If user see this page shouldn't it have gone through `_document` first ?

XavierJp: As for the other types(non-agents), they are not our focus right now so I would not add this only for them

pages/formulaire/nps.tsx
```diff
@@ -16,7 +20,7 @@ export const visitorTypes = [
     label: 'Dirigeant(e) d’entreprise ou d’association',
   },
   {
-    value: 'Agent public',
+    value: 'Agent public non connecté',
     label: 'Agent public',
   },
```
XavierJp, line 25: If we change it we should also change https://github.com/etalab/annuaire-entreprises-site/blob/1b8eeedee168cd46c77c0c202d56725eb459ce0d/clients/matomo/index.ts#L234-L235

to take into account every userType starting with `agent` to be considered as Agents Publics in the stats other wise it will break history

XavierJp (approved):
We also talked about reseting NPS banner display/hide every 6 months. Sorry I forgot to add it in the issue. Dont know if you want to tackle it here or in another PR

Overall : 
- I am not sure about the custom dimensions for the unauthencated users as it is very limited (few hundreds per month and not our current focus)
- form update is neat 👍 
- session in the props might raise security issue for any MITM attack

pages/_document.tsx
johangirod: We already serialize the whole session object in the `NEXT_DATA` props for every page with react enabled. The change here doesn't modify the session object, nor the surface of information accessible from the front, only read in it. 

Unless there is something I missed?

pages/formulaire/merci.tsx
johangirod: Yes, I made this change to start collecting behavior from other user types, for us to have informed data when we decide to improve UX for them as well (instead of having to wait for a specific research campaign). But I'm also understanding the « keep it minimum » approach when it comes to user tracking, so I can remove this.

XavierJp (approved):

XavierJp:
I wait for agent connect debugging to merge this

XavierJp:
@johangirod ready to merge this but can you please first solve the merge conflict ?

XavierJp (approved):

PR #856, opened 2024-01-09
feat: improve elus section

### LINKED ISSUE #549: Présentation de la liste des élus d'une collectivité
Quand je "parse" visuellement les informations affichées dans le tableau présentant la liste des élus d'une collectivité, j'ai l'impression que la colonne "rôle", qui n'est pas présente sur toutes les lignes, forme des groupes :

image

Comme si, dans la capture ci-dessus, il y avait 6 personnes dans la catégorie "3ème adjoint au maire". Or ce n'est évidemment pas comme cela qu'il faut comprendre ce tableau, il n'y a qu'un 3ème adjoint et les noms en dessous ont un rôle non renseigné.

Idées pour corriger ce problème :
- Mettre la colonne "rôle" en deuxième après la colonne "nom"
- ajouter un rôle "inconnu" / "non renseigné" (peut-être un peu lourd)
- avoir une séparation visuelle entre les lignes pour forcer l'interprétation ligne par ligne (fond zébré ou bordures)

johangirod:
- sort by role (maire first, then by adjoint number, then by name)
- display date of birth separately
- add « non renseigné » when role is not provided

image

Considering doing the same for `dirigeant` section, what do you think?

closes #549

### DIFF
--- a/components/dirigeants-section/elus-section.tsx
+++ b/components/dirigeants-section/elus-section.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
+import NonRenseigne from '#components/non-renseigne';
 import { EAdministration } from '#models/administrations/EAdministration';
 import { IEtatCivil } from '#models/immatriculation';
 import { isCollectiviteTerritoriale, IUniteLegale } from '#models/index';
-import { formatDatePartial } from '#utils/helpers';
+import { capitalize, formatDatePartial } from '#utils/helpers';
 import { Section } from '../section';
 import { FullTable } from '../table/full';
 
@@ -26,10 +27,12 @@ const ElusSection: React.FC<{ uniteLegale: IUniteLegale }> = ({
     }${(elu.nom || '').toUpperCase()}`;
 
     const infos = [
-      elu.role ?? <span style={{ color: '#555' }}>Non renseigné</span>,
+      elu.role ?? <NonRenseigne />,
       <>{nomComplet}</>,
-      <span style={{ textTransform: 'capitalize' }}>
-        {formatDatePartial(elu.dateNaissancePartial)}
+      <span>
+        {capitalize(formatDatePartial(elu.dateNaissancePartial) ?? '') || (
+          <NonRenseigne />
+        )}
       </span>,
     ];
 
@@ -63,7 +66,7 @@ const ElusSection: React.FC<{ uniteLegale: IUniteLegale }> = ({
           </p>
         )}
       </Section>
-      <style global jsx>{`
+      <style jsx>{`
         table > tbody > tr > td {
           min-width: 30%;
         }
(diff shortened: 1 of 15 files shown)


components/dirigeants-section/elus-section.tsx
```diff
@@ -65,11 +64,38 @@ const ElusSection: React.FC<{ uniteLegale: IUniteLegale }> = ({
         )}
       </Section>
       <style global jsx>{`
-        table > tbody > tr > td:first-of-type {
-          width: 30%;
+        table > tbody > tr > td {
+          min-width: 30%;
         }
       `}</style>
```
XavierJp, line 70: Hum this is dangerous as there is a `global` keyword so the style will apply to every `table > tbody > tr > td`

This was poorly placed in the first place (my fault!) but you should check if this does not break table with multiple columns elsewhere

components/dirigeants-section/elus-section.tsx
```diff
@@ -26,10 +26,11 @@ const ElusSection: React.FC<{ uniteLegale: IUniteLegale }> = ({
...
     const infos = [
-      elu.role,
-      <>
-        {nomComplet}, né(e) en {formatDatePartial(elu.dateNaissancePartial)}
-      </>,
+      elu.role ?? <span style={{ color: '#555' }}>Non renseigné</span>,
+      <>{nomComplet}</>,
+      <span style={{ textTransform: 'capitalize' }}>
+        {formatDatePartial(elu.dateNaissancePartial)}
+      </span>,
```
XavierJp, line 33: There is a js helper `capitalize()` if you want to avoid inline style (https://github.com/etalab/annuaire-entreprises-site/blob/1b8eeedee168cd46c77c0c202d56725eb459ce0d/utils/helpers/formatting/formatting.ts#L164)

Maybe that `#555` should be color placed in a class in `global.css` rather than repeated in every line ?

Also `Non renseigné` is not coloured elsewhere in the app, only italic. I think for coherence we can :
- either only use italic here
- either generalize grey to evey `Non renseigné` in the app

components/dirigeants-section/elus-section.tsx
```diff
@@ -65,11 +64,38 @@ const ElusSection: React.FC<{ uniteLegale: IUniteLegale }> = ({
...
+    return 1;
+  }
+  if (roleB == null) {
+    return -1;
+  }
+  if (roleA.match(/^[\d]+/) && roleB.match(/^[\d]+/)) {
+    return parseInt(roleA, 10) < parseInt(roleB, 10) ? -1 : 1;
+  }
+  return roleA < roleB ? -1 : 1;
+}
```
XavierJp, line 101: 👍

XavierJp (approved):

XavierJp:
And I think it is an excellent idea to apply this to the dirigeant section (and maybe beneficiaire too)

PR #864, opened 2024-01-22
refactor: migrate outdated <b/> in favor of <strong/>

XavierJp:
See https://developer.mozilla.org/fr/docs/web/HTML/Element/strong#b_ou_strong

And #854


johangirod (approved):
:clap: Quelle rapidité !

PR #881, opened 2024-01-29
fix: text color in label in search page

### LINKED ISSUE #875: [BUG] Regression visuelle sur les label dans la recherche

johangirod:
fix #875


XavierJp (approved):
Beautiful

PR #887, opened 2024-01-30
refactor: uses recherche entreprise for all etablissement

XavierJp:
## Problem 
Currently `/etablissement` defaults on API Sirene

## Solution
This is a proposition to apply the same logic as unite legale but for etablissements

exemple of an etablissement with an IDCC : 55203253400646

Tricky part : we still need to call both Insee et Recherche API as recherche does not tells us if etablissement is non-diffusible (yet)


models/etablissement.ts
```diff
@@ -24,31 +33,107 @@ const getEtablissementFromSlug = async (
...
+  const etablissement = fetchFromClients(siret, isBot);
 
   return etablissement;
 };
 
 /**
  * Return an Etablissement for a given siret
  */
-const getEtablissement = async (siret: Siret): Promise<IEtablissement> => {
+const fetchFromClients = async (
```
johangirod, line 45: It seems to me that the logic for orchestrating the various calls is exactly the same as for `unite-legale.ts`. Could we factor this into a function? Something like : 

```typescript
function fetchFromInseeOrchestrator<IRecherche | IInsee>(
   isBot: boolean, 
   apiRecherche: () => Promise<IRecherche | IAPINotResponding>,
   insee (useFallback: boolean) => Promise<IInsee | IAPINotResponding>,
) : Promise<IRecherche | IInsee | IAPINotResponding>

File: models/etablissement.ts [hunk 2]
```diff
@@ -24,31 +33,107 @@ const getEtablissementFromSlug = async (
...
+
+  if (rechercheEntrepriseFailed) {
+    return true;
+  } else {
+    if (isBot) {
+      return false;
+    }
+
+    // we always call insee for etablissement as we dont know if they belong to an EI
+    return true;
```
johangirod, line 122: We don't know, but we could know, as the SearchAPI does return if it's an EI or not. We could : 
- Adapt the `clientEtablissementRechercheEntreprise` client to return this information
- Adapt the model to call for `uniteLegale` (with the cache logic, it should not do more than one call).

models/etablissement.ts
```diff
@@ -61,48 +146,29 @@ const getEtablissement = async (siret: Siret): Promise<IEtablissement> => {
...
-          administration: EAdministration.INSEE,
-          context: {
-            siret,
-          },
-        });
-        logFatalErrorInSentry(error);
-        throw error;
-      }
-    }
+    return APINotRespondingFactory(EAdministration.INSEE, 500);
```
johangirod, line 171:
```suggestion
    return APINotRespondingFactory(EAdministration.DINUM, 500);
```

models/etablissement.ts
johangirod: Either way, we could add a `isEi` callback to the orchestrator.

XavierJp:
@johangirod I managed to add `estEntrepreneurIndividuel` into `EtablissementComplements` and I think it turned out pretty well. No additional call needed.

I am still trying to refacto `fetchFromInseeOrchestrator`

models/core/etablissement.ts
```diff
@@ -52,7 +53,11 @@ const fetchFromClients = async (
   const etablissementRechercheEntreprise =
     await fetchEtablissementFromRechercheEntreprise(siret, useCache);
 
-  const useInsee = shouldUseInsee(etablissementRechercheEntreprise, isBot);
+  const useInsee = shouldUseInsee(
```
johangirod, line 56: Nice to have this factorised in a separate function :)

HAEKADI (approved):

PR #893, opened 2024-02-06
Ask for NPS every 6 month

johangirod:
WIP


frontend/src/helpers-feedbacks.js
```diff
@@ -16,24 +16,49 @@ function init() {
...
+        var currentTime = new Date();
+        var diffInMonths =
+          (currentTime - lastTimeDate) / (1000 * 60 * 60 * 24 * 30);
+        if (diffInMonths < periodicity) {
+          return;
+        }
+      }
+    } catch (e) {}
+  }
+
```
XavierJp, line 61: I would suggest this implementation :

- I find it a bit simpler but this is very personnal
- to avoid conflict with previous localStorage key I would change the key to "nps-modal-2"
- I think using timestamp might be a bit safer but it might be my old js `Date()` trauma speaking !

```
function triggerModal(
  modalId,
  triggerCount = 1,
  startsWithString = '/',
  monthPeriodReset = 6
) {
  var path = window.location.pathname;

  try {
    var lastTriggerTimestamp = window.localStorage.getItem(modalId);
    var diffInMonths =
      (new Date().getTime() - lastTriggerTimestamp) / (1000 * 60 * 60 * 24 * 30);
    if (diffInMonths > monthPeriodReset) {
      // after X month, reset
      window.localStorage.removeItem(modalId);
    }
  } catch (e) {
    // no access to local storage. Probably for privacy reason. We choose not to show modal rather than showing it on every visit
    return;
  }

  if (path.indexOf(startsWithString) === -1) {
    return;
  }

  // page-view counter increment
  var pvKey = 'pv-' + modalId;
  var pageViewCount = window.sessionStorage.getItem(pvKey) || 0;

  if (pageViewCount >= triggerCount) {
    window.showModal(modalId);
  }

  window.sessionStorage.setItem(pvKey, parseInt(pageViewCount, 10) + 1);
}
```

johangirod: Perfect, looks clearer to me too ! Just not sure what happens when `lastTriggerTimestamp` is undefined the first time the function is called. The other thing is that it will show modal for everyone, even those who just clicked on « don't show me this again » recently. 

If we're ok with that, I'm all in favor of your version.

XavierJp: Thanks I have to fix the. initialization. 

I think we are ok with the dont show me again being display again.

XavierJp (approved):

PR #918, opened 2024-02-21
style: generalize #fef6e3 as only light yellow accross site

XavierJp:
Use the same yellow as the agent illu for the NPS banner and for connexion pages' background

Not sure about the connexion background

What do you think ?


renardpal:
i'd say nice for the nps, keep it blue for the background but is this page still really used?

XavierJp:
Connexion page not really but the connexion error will still be used

johangirod (approved):

PR #919, opened 2024-02-21
Pasta salad

XavierJp:
Not best practise, but this is a mix of fixes. Each commit can be read indendantly

- change the way we log server error in order to be able to read them in Kibana
- handle >4MB actes & bilans
  - we should probably move to a dedicted service that would download files and handles queues
  - @MKCG 
- date fermeture
- recipient siret to API Entreprise

### DIFF
--- a/components/unite-legale-section/index.tsx
+++ b/components/unite-legale-section/index.tsx
@@ -87,7 +87,14 @@ const UniteLegaleSection: React.FC<{
         ]
       : []),
     ...(!estActif(uniteLegale)
-      ? [['Date de fermeture', formatDate(uniteLegale.dateDebutActivite)]]
+      ? [
+          [
+            'Date de fermeture',
+            formatDate(
+              uniteLegale.dateDebutActivite || uniteLegale.dateFermeture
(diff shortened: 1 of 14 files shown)


XavierJp:
https://nextjs.org/docs/messages/api-routes-response-size-limit#possible-ways-to-fix-it

components/unite-legale-section/index.tsx
```diff
@@ -87,7 +87,14 @@ const UniteLegaleSection: React.FC<{
         ]
       : []),
     ...(!estActif(uniteLegale)
-      ? [['Date de fermeture', formatDate(uniteLegale.dateDebutActivite)]]
+      ? [
+          [
+            'Date de fermeture',
+            formatDate(
+              uniteLegale.dateDebutActivite || uniteLegale.dateFermeture
```
johangirod, line 94: Très chelou à lire, mais j'imagine que c'est le comportement correct (j'aurais eu tendance à inverser les termes du `ou`).

johangirod (approved):
Top !

PR #938, opened 2024-03-12
895 various fixes

johangirod:


components-ui/button/index.tsx
```diff
@@ -1,8 +1,7 @@
 'use client';
```
XavierJp, line 1: No need to use it as a client component anymore ?

utils/server-side-helper/app/error-page.ts
XavierJp, line 1: Weirdly this appear as an empty file on github

XavierJp (approved):

components-ui/button/index.tsx
johangirod: I think we do, as there is an event callback forwarding (onClick). I'll check on that.

utils/server-side-helper/app/error-page.ts
johangirod: I think I pushed the empty file, my bad.

components-ui/button/index.tsx
johangirod: Confirmed here: https://react.dev/reference/react/use-client#limitations

PR #955, opened 2024-03-18
[Agents] Module feedback / recrutement

johangirod:


XavierJp (approved):
Few issues : 
- You forgot to change `<Question />` in the `/page` router
- I find the purple button a bit agressive. Sounds like an error
- I would prefer that `Une question ?` and `Agent Feedback` have the same behaviour. currently they are not consistent on hover
- Can we align modal styles with `<AgentWall />` (again for consistency) ?
- Can we align multi select style with `<MultiChoice />` ? I feel like it is doing the same thing, it should look the same (or we could even use the same component)
- Beta subscription message implies we are going to make a visio with every person that volounteer. Might a bit risky as they will expect a call and we already have 4000 accounts.

Sounds very nice overall

johangirod:
> You forgot to change <Question /> in the /page router

It's because you can't have server action in page routes. However, we use them for submitting agent feedback. Thus, I choose to display the previous button, and wait for them to be migrated to the new architecture.

Totally agree with the rest of the issues, I'm working on it :+1:

XavierJp:
FYI There is an issue with CRISP at the moment

XavierJp:
Maybe : 
- use a 2px lighter border
- use the same `Fermer X` as search filter in mobile
- add a listener outside modale to close ? (not sure about this one)
- merge with main and syncro with button
- use buit i form verfiication rather than button deactivation

components-ui/filter-menu/filter-menu.tsx
```diff
@@ -117,7 +127,7 @@ export const FilterMenu: React.FC<PropsWithChildren<FilterMenuProps>> = ({
             background-color: #fefefe;
           }
 
-          label.close-container {
+          .close-container {
```
XavierJp, line 130: Remove close-container

XavierJp (approved):

PR #959, opened 2024-03-25
feat: dysyncronize view from user rights

XavierJp:
references #945


XavierJp:
Closing and repopening to trigger scalingo deployement

XavierJp:
Funny how it really does not work

XavierJp:
Oh actually it did

HAEKADI (approved):
I don't really know what I'm approving 😅

XavierJp:
Sorry @HAEKADI, this PR is a refacto that tries to uncouple view and model. 

we use to have complexe conditions in the view every time we wanted to displayed agent-only data. With this PR we centralize it as scopes that can be tested from the view. 

What used to be multiple `isAgent(session)` is now centralized into a string that defines what can or cannot see an agent.

PR #970, opened 2024-04-02
fix: remove non diffusible document for agent

johangirod:


XavierJp (changes requested):

XavierJp (approved):
Need to fix build first but seems good to me

PR #997, opened 2024-04-12
feat: migrate carte to /app and improve layout

XavierJp:
- solve CSS issues with search filter
- solve header issues (missing searchterm, session etc.)
- migration from /rechercher to /app router
- improve etablissement pointers

What is out of scope :
- add hover behaviour on map markers and search results


app/(header-search)/rechercher/page.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const searchFilterParamsJSON = searchFilterParams.toJSON();
+
+  return (
+    <>
+      <NPSBanner />
+      <HeaderWithAdvancedSearch
+        useSearchBar={true}
+        useAgentCTA={true}
+        useMap={false}
+        searchParams={searchFilterParamsJSON}
```
johangirod, line 48: On a du JSON d'un côté et du `IParams` attendu de l'autre, mais pas d'erreur typescript ?
Ah, je viens de comprendre. La méthode toJSON est très mal nommé, il s'agit d'avantage d'un `toSerializableObject`

components/dirigeants-section/is-personne-morale.ts
```diff
@@ -0,0 +1,18 @@
+import { IEtatCivil, IPersonneMorale } from '#models/immatriculation';
+
+/**
+ * Weird bug happennig here. Webpack build fail when this function is in model/dirigeants.ts
+ * @param toBeDetermined
+ * @returns
+ */
```
johangirod, line 7: Might be because of a circular dependancy.

johangirod (commented):
This is a good thing to use proper import of maplibre-gl library inside react instead of using an inserted script. 

However, it appears that there is still a lot of error due to this map logic : 
https://errors.data.gouv.fr/organizations/sentry/issues/?groupStatsPeriod=auto&page=0&query=maplibre&referrer=issue-list&statsPeriod=14d

We might want to fix thoses once and for all at some point.

app/(header-search)/rechercher/page.tsx
XavierJp: Yes, bien vu

PR #1009, opened 2024-04-22
fix: flickering of loading state for datasection component

johangirod:


XavierJp (approved):
If I understand well we render nothing during the first 100ms. For the record do we at least render an empty section (with `<hn>` and `source`) ?

johangirod:
@XavierJp, we render an empty section with 100px height. You can find that logic is here :
https://github.com/etalab/annuaire-entreprises-site/pull/1009/files#diff-10db91f863b69b4afc7dd58fa7d4760bea85dc3336843314356a7c5dc44787b5L36-L46

PR #1010, opened 2024-04-22
957 fix copy safari edge

johangirod:


components/table/copy-paste.tsx
```diff
@@ -82,11 +94,13 @@ export function CopyPaste({
...
             <>
-              <CheckMarkSVG /> copié
+              <CheckMarkSVG />
+              <span> copié</span>
             </>
           ) : (
             <>
-              <CopySVG /> copier
+              <CopySVG />
+              <span> copier</span>
```
johangirod, line 103: This is the important bit. We need to enclose conditionally rendered `TextNode` inside an element for the translation to work.

XavierJp (approved):

PR #1013, opened 2024-04-23
Add OpenData from API Entreprise for agent

johangirod:
- [x] Add a new scope `opendata`
- [x] Add EORI client & model
- [x] Add carte professionnelle des travaux publics
- [x] Add Qualibat
- [x] Add Qualifelec
- [x] Add Opqibi
- [x] Date de naissance
- [x] Add link & label from résumé to certification for agent 
- [x] Add E2E test to check that standard users don't get to see protected data


models/user/rights.ts
```diff
@@ -31,6 +45,27 @@ export function hasRights(session: ISession | null, rightScope: EScope) {
...
+
 export function isLoggedIn(session: ISession | null) {
   return (session?.user?.scopes || []).length > 0;
 }
+
+export type INotAuthorized = {
+  __I_NOT_AUTHORIZED__: never;
+};
+
+export function isNotAuthorized<T>(
```
johangirod, line 60: We add a new base type for handling not authorized logic directly in the model when it makes sense (and not doing it systematically in the view).

johangirod (commented):
Ready for review

XavierJp:
I like the idea of IUnauthorized but I dont think it is working : 
- you end up with a lot more session being pass as arguments
- you still need to handle IUnauthorized in view as
- you will still need to have a way for the view to prevent API call based on rights. See conformite for instance. Otherwise it will lead to many unnecessary calls
- As API calls from API Entreprise tend to be very slow (appart from the test env) we will need to use async a lot

XavierJp:
I like the see more, but we need something much more obvious visually. I missed it several time in local trying to force trigger it, before eventually I realised I had it in front of me, I just did not see it.

Maybe github can be an inspiration on this, I find that their collapsed comment component is quite explicit

johangirod:
> every call should be made async as API Entreprise tend to be much slower than sirene and the whole point of the /app router migration was to allow more async call from /entreprise page

This is only the case for agent. If you believe it's too much of trade off, we could use async server component. This would enable us to load the page first and then render, without having to move calls to an API route first.

> require to share the session to the model, which I tried to avoid

I think it is the point. The view shouldn't have to know which API call to make depending on authorization level. This is a specific domain logic that should be inside the model. If you think sharing the session with the model is not a good pattern (I do agree, but I did it for the sake of simplicity :p), we can refactor, and only share `the User` model, or just the `scopes`.

> see more is a really good idea, but this implementation is not obvious enough (I add a really hard time finding it, knowing it was there) 

We cannot do without because otherwise, the `carteProfessionnelle` document would be pushed down very far on the page, especially when there is a lot of actes (like it does with large companies). We can however fix it's visibility and affordability.

> we should move them into a dedicated section, not replace the rne section, even if it creates duplicates as RNE and RCS are not the same and dont hold the same meaning.

From a data perspective, I agree, but does-it from a user-centric standpoint ? I'm only asking, to understand better. In which case the RCS would differ from the RNE ?

>  we need a way to display badges on /entreprise page and/or /etablissement page (if they make sense at siret level)

Yes !

app/(header-default)/dirigeants/_component/sections/mandataires-rcs.tsx
```diff
@@ -25,76 +25,68 @@ type IProps = {
...
- * Dirigeants section
+ * Dirigeants from RCS section
  */
 const MandatairesRCSSection: React.FC<IProps> = ({
   mandatairesRCS,
   immatriculationRNE,
   uniteLegale,
 }) => (
-  <Section
+  <DataSectionClient
```
johangirod, line 35: `ServerSection` me parait suffisant - et necessaire ici, non ?

XavierJp (approved):

app/(header-default)/dirigeants/_component/sections/mandataires-rcs.tsx
XavierJp: En effet bien vu !

.env.dev
```diff
@@ -22,7 +22,7 @@ AGENTCONNECT_URL_DISCOVER=https://fca.integ01.dev-agentconnect.fr/api/v2/.well-k
...
...wX2RlY2xhcmFudDJfZGF0ZV9uYWlzc2FuY2UiLCJkZ2ZpcF9kYXRlX3JlY291dnJlbWVudCIsImRnZmlwX2RhdGVfZXRhYmxpc3NlbWVudCIsImRnZmlwX2FkcmVzc2VfZmlzY2FsZV90YXhhdGlvbiIsImRnZmlwX2FkcmVzc2VfZmlzY2FsZV9hbm5lZSIsImRnZmlwX25vbWJyZV9wYXJ0cyIsImRnZmlwX25vbWJyZV9wZXJzb25uZXNfYV9jaGFyZ2UiLCJkZ2ZpcF9zaXR1YXRpb25fZmFtaWxpYWxlIiwiZGdmaXBfcmV2ZW51X2JydXRfZ2xvYmFsIiwiZGdmaXBfcmV2ZW51X2ltcG9zYWJsZSIsImRnZmlwX2ltcG90X3JldmVudV9uZXRfYXZhbnRfY29ycmVjdGlvbnMiLCJkZ2ZpcF9tb250YW50X2ltcG90IiwiZGdmaXBfcmV2ZW51X2Zpc2NhbF9yZWZlcmVuY2UiLCJkZ2ZpcF9hbm5lZV9pbXBvdCIsImRnZmlwX2FubmVlX3JldmVudXMiLCJkZ2ZpcF9lcnJldXJfY29ycmVjdGlmIiwiZGdmaXBfc2l0dWF0aW9uX3BhcnRpZWxsZSIsIm1lc3JpX2lkZW50aWZpYW50IiwibWVzcmlfaWRlbnRpdGUiLCJtZXNyaV9pbnNjcmlwdGlvbl9ldHVkaWFudCIsIm1lc3JpX2luc2NyaXB0aW9uX2F1dHJlIiwibWVzcmlfYWRtaXNzaW9uIiwibWVzcmlfZXRhYmxpc3NlbWVudHMiLCJwb2xlX2VtcGxvaV9pZGVudGl0ZSIsInBvbGVfZW1wbG9pX2FkcmVzc2UiLCJwb2xlX2VtcGxvaV9jb250YWN0IiwicG9sZV9lbXBsb2lfaW5zY3JpcHRpb24iLCJwb2xlX2VtcGxvaV9wYWllbWVudHMiLCJtZW5fc3RhdHV0X3Njb2xhcml0ZSIsIm1lbl9zdGF0dXRfYm91cnNpZXIiLCJtZW5fZWNoZWxvbl9ib3Vyc2UiXSwic3ViIjoic3RhZ2luZyBkZXZlbG9wbWVudCIsImlhdCI6MTY5MzkwNTAyNCwidmVyc2lvbiI6IjEuMCIsImV4cCI6MjAwOTUyNDIyNH0.uKkMeXNmzwaultKAuS6l1o9StrZky-mY7XLTzygdut4
```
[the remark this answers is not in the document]
XavierJp: We havent add GitGuardian yet

PR #1028, opened 2024-04-29
Monitoring : Log des accès effectué par les agents

MKCG:
Closes : https://github.com/etalab/annuaire-entreprises-site/issues/976

- [ x] siret des orgas de l’usager (ex DINUM)
- [ x] email de l'usager ou slug agent connect (NB : email agent = donnée publique)
- [ x] route consultée sur API Entreprise
- [   ] cas d'usage juridique (I don't know what should be done here)
- [ x] timestamp
- [ x] Archivage : 3 mois (https://github.com/etalab/annuaire-entreprises-infrastructure/pull/123)


XavierJp:
@MKCG I refactored your code a bit, made generic in order to be able to log any request, not just API Entreprise's

I have a few questions about architecture though : 
- the route is already logged by the default logger - why the need to have an additionnal logger ? 
- we can also modify the logger and add siret, email and scopes when calls are made by an agent 
- can https://github.com/etalab/annuaire-entreprises-infrastructure/pull/123 post process the logs before ingesting it ?

MKCG:
> @MKCG I refactored your code a bit, made it more generic in order to be able to log any request, not just API Entreprise's
> 
> I have a few questions about architecture though :
> 
>     * the route is already logged by the default logger - is their really a the need for an additionnal logger ?
> 
>     * we can also modify the default logger and add siret, email and scopes whenever calls are made by an agent.
> 
>     * current implementation require us to add call. I would prefer an implementation built in `httpClient`
> 
>     * can [Monitoring : store logs from authenticated agents into a dedicated bucket and backup them at least once per day annuaire-entreprises-infrastructure#123](https://github.com/etalab/annuaire-entreprises-infrastructure/pull/123) post process the logs before ingesting it ?

I don't know how to retrieve the current session from the logInterceptor, I have this error when I try to import it : 
image

image

johangirod:
> we can also modify the default logger and add siret, email and scopes whenever calls are made by an agent.

Should we, thought? It means that the email & Siret will be logged everywhere. If we are in the mindset of logging only what we need, this is a bit overkill. I'm comfortable having a specific logging logic for API Entreprise, as it comes from a specific need.

On the other hand, it would allow us to have more stats about usage of external API.

> I don't know how to retrieve the current session from the logInterceptor, I have this error when I try to import it

With Nextjs you have two different ways to render pages that use different API for retrieving cookies. It might work by setting the session info into sentry `scope` object when deserializing session. I can look into it if we go choose to go that way.

XavierJp (approved):

clients/api-entreprise/client.ts
```diff
@@ -30,6 +31,8 @@ export default async function clientAPIEntreprise<T, U>(
     throw new HttpUnauthorizedError('Missing API Entreprise credentials');
   }
 
+  sensitiveRequestLogger(route);
```
johangirod, line 34: Il manque pas un await ici ?

johangirod (commented):
Ça vient peut-être d'un await manquant ? @MKCG

PR #1046, opened 2024-05-13
refactor: api route factorization

### LINKED ISSUE #1045: [ENV] S'assurer que le MATOMO_SITE_ID est bien en variable d'env dans la récupération des stats
Sur le site il est intégré aux urls

XavierJp:
closes #1045 1045

refacto API routes and can be useful for #1013 

Note this PR will have consequences : 

Matomo will now be activated in production build in local and staging

@johangirod should we add another env variabel `MATOMO_ENABLED` that deactivate matomo client in local and staging ?


johangirod (approved):
Top ! The factorisation of API route is clean and easy.

johangirod:
I think `MATOMO_ENABLED` is a good idea; otherwise the stat page would fail in dev & staging. I don't really see how we could do otherwise.

PR #1047, opened 2024-05-14
Use server component instead of client data fetching

johangirod:


app/(header-default)/dirigeants/_component/sections/index.tsx
```diff
@@ -1,60 +1,53 @@
...
+export async function DirigeantInformation({
   uniteLegale,
-  mandatairesRCS,
+  session,
 }: {
   uniteLegale: IUniteLegale;
-  mandatairesRCS: Array<IDirigeant> | IAPINotRespondingError | null;
+  session: ISession | null;
 }) {
-  const immatriculationRNE = useFetchImmatriculationRNE(uniteLegale);
```
XavierJp, line 27: Idem we can delete `useFetchImmatriculationRNE`

PR #1052, opened 2024-05-14
Move association and protected certificat to server component

johangirod:


app/api/data-fetching/association/[slug]/route.ts
Comment (XavierJp) [inline app/api/data-fetching/association/[slug]/route.ts:1]: doesnt this introduce a regression as we dont have `withIgnoreBot`anymore ?
Asso API is very rate limited and we cant trigger a call for bot

components-ui/alerts-with-explanations/association-adress.tsx
```diff
@@ -10,15 +10,14 @@ import {
...
 import { Warning } from '../alerts';
 
 const AssociationAdressAlert: React.FC<{
   uniteLegale: IAssociation;
   association: IDataAssociation | IAPINotRespondingError | null;
-}> = ({ uniteLegale, association }) => {
-  const session = useSession();
-
+  session: ISession | null;
+}> = ({ uniteLegale, association, session }) => {
```
XavierJp, line 20: session is imported but not used ?

app/(header-default)/labels-certificats/[slug]/page.tsx
```diff
@@ -97,8 +83,7 @@ const LabelsAndCertificatsPage = async (props: AppRouterProps) => {
           session={session}
         />
         {!checkHasLabelsAndCertificates(uniteLegale) &&
-          (!protectedCertificats ||
-            !hasProtectedCertificatsEntreprise(protectedCertificats)) && (
+          !hasRights(session, EScope.protectedCertificats) && (
```
Comment (XavierJp) [inline app/(header-default)/labels-certificats/[slug]/page.tsx:86]: I feel like these conditions dont work as we will miss the `<p>Cette structure ne possède aucun label ou certificat.</p>` in case we have `EScope.protectedCertificats` but no answer for protectedCertificats

Then
- either we do a full server render for this page, like dirigeant
- either we use API routes as they seem more relevant to this multiple calls architecture

XavierJp (commented):
I have mixed feelings. I feel like this new architecture introduce minor regressions for the sake of writing easier code

Resulting code is much simpler and I love that. But maybe association & ProtectedCertif are not the best use case for this.

Association introduce :
- a potention regression on our SEO (? this has yet to be proved or debunked) 
- less accurate bot management (this can/should be fixed)

Protected Certif introduce :
- issue of how you display the no certificate message

I have the feeling that server side rendering is really good the way you did it on `/dirigeant` page otherwise every interdependant component like the `<p>Cette structure ne possède aucun label ou certificat.</p>` will be a headache

app/api/data-fetching/association/[slug]/route.ts
Comment (johangirod) [reply app/api/data-fetching/association/[slug]/route.ts:1]: Removed call for API & RNE for bots here : https://github.com/annuaire-entreprises-data-gouv-fr/site/pull/1054/commits/b74e19d89e836e650c22cf29e16fd2f8ba9b47ec

PR #1069, opened 2024-05-22
Various sentry error fixes

johangirod:
- Hydration error
- Grouping error on sentry
- NotAllowedError
- ...


XavierJp (approved):

XavierJp:
The change on UI component can be breaking (`div` > `span`), this need to be extensively tested once deployed to ensure no regression get introduce

johangirod:
> The change on UI component can be breaking (div > span)

The `display` property of the two elements is already forced in the css, so this should not be breaking.

PR #1082, opened 2024-05-29
fix: only rely on DGEFP for qualiopi certification

XavierJp:
In addition to more frequent dataset update on recherche-entreprise side, this PR made a change in Organisme de Formation section. 

We now only rely on DGEFP response to determine if an organisme is Qualiopi as recherche entreprise can be outdated.

This mean we loose the ability to have "certifié qualiopi" in the section title, but this seem like a fair trade off.

We will still have outdated data in summary and in the search engine so this will not entirely solve the issue, but it will improve it.


HAEKADI (approved):
Btw after the daily data update, the data will have at worst a 24 hours lag.

XavierJp:
> Btw after the daily data update, the data will have at worst a 24 hours lag.

And this is much better, congrats !

PR #1096, opened 2024-06-04
feat: use Link in tabs

XavierJp:
references #1066


XavierJp:
And Scalingo is BACK !

@johangirod it was indeed related to the orga migration

PR #1103, opened 2024-06-10
feat: add a KBIS link in etablissement page as some user use siret

XavierJp:
Quick improvement upon siret page as some users use siret and dont find the RNE/KBIS


HAEKADI (approved):
Good idea!

PR #1104, opened 2024-06-10
refactor: factorise code from API transport layer

johangirod:
Todo : 
- [x] Validate the proposition (@XavierJp)
- [x] Extend to all API routes
- [x] Add a `IAPITransportLayerError` type returned by front end
- [x] Handle Unautorized & DataFetching error inside `DataSection` component
- [x] Differentiate between network error (client side) and model error (server side) in `useFetchAPIRoute`

### Screenshot

En cas d'erreur dans la couche transport : 
image


app/api/data-fetching/index.ts
XavierJp, line 1: Empty file ?

app/api/data-fetching/routes.ts
```diff
@@ -0,0 +1,51 @@
...
+export const APIRoutesScope: Record<APIPath, false | EScope> = {
+  'espace-agent/carte-professionnelle-TP':
+    EScope.carteProfessionnelleTravauxPublics,
+  'espace-agent/conformite': EScope.conformite,
+  'espace-agent/opqibi': EScope.protectedCertificats,
+  'espace-agent/qualibat': EScope.protectedCertificats,
+  'espace-agent/qualifelec': EScope.protectedCertificats,
+  'espace-agent/rcs-mandataires': EScope.mandatairesRCS,
+  'espace-agent/rne/documents': EScope.documentsRne,
+  rne: false,
```
XavierJp, line 35: Could it make sense to merge `APIRoutesScope` and `APIHandler` ojects ?

Something in the likes : 
```
{
	[route:string] : {
		handler: APIHandler
		scope: EScope
	}
}
```

app/api/data-fetching/routes.ts
```diff
@@ -0,0 +1,51 @@
...
+  rne: NotImplementedHandler,
+  association: NotImplementedHandler,
+  'verify-tva': NotImplementedHandler,
+  geo: NotImplementedHandler,
+  stats: NotImplementedHandler,
+} as const;
+
+export type APIPath = keyof typeof APIHandler;
+
+export const APIRoutesScope: Record<APIPath, false | EScope> = {
```
XavierJp, line 26: Why `false` over `null` ?

app/api/data-fetching/routes.ts
XavierJp, line 1: `routes.ts` is pretty close to `route.ts` which is reserved by next. I would advise to use a more explicitely different name here

app/api/data-fetching/[...slug]/route.ts
```diff
@@ -0,0 +1,82 @@
...
+  { params }: { params: { slug: Array<string> } },
+  /* The `session` parameter in the `getRoute` function is an object of type `ISession` representing
+  the user session. It contains information about the current user's session, such as user details,
+  permissions, and any other relevant data needed to process the request. The `session` object is
+  used to check if the user has the necessary rights or scopes to access a specific API route before
+  handling the request. */
+  session: ISession
+) {
+  const slug = params.slug.at(-1) as string;
+  const route = params.slug.slice(0, -1).join('/');
```
Comment (XavierJp) [inline app/api/data-fetching/[...slug]/route.ts:24]: This would be nice in a helper as it is done twice in the file

I would maybe add sentry logs as it looks bug prone

app/api/data-fetching/routes.ts
XavierJp: Or why not use a dedicated scope `EScope.none` for open routes ? It would ensure type coherence

app/api/data-fetching/[...slug]/route.ts
```diff
@@ -0,0 +1,82 @@
...
+  const scope = APIRoutesScope[route as APIPath];
+  let response;
+  if (scope) {
+    if (!hasRights(session, scope)) {
+      if (session?.user?.scopes) {
+        throw new HttpUnauthorizedError('Scope not found for API route');
+      }
+      throw new HttpForbiddenError(
+        'User does not have the required scope for this API route'
+      );
```
Comment (XavierJp) [inline app/api/data-fetching/[...slug]/route.ts:38]: According to me both are a `HttpUnauthorizedError`

app/api/data-fetching/[...slug]/route.ts
```diff
@@ -0,0 +1,82 @@
...
+        // error, as it clearly indicates that the error comes from the application.
+        logInfoInSentry(
+          new Exception({
+            name: 'APIRouteError',
+            cause: e,
+            context: { page: route, slug },
+          })
+        );
+        return new Response(e.message, { status: e.status });
+      }
```
Comment (XavierJp) [inline app/api/data-fetching/[...slug]/route.ts:70]: I dont fully understand this part

app/(header-default)/documents/[slug]/_components/carte-professionnelle-TP-section.tsx
```diff
@@ -5,15 +5,17 @@ import { Icon } from '#components-ui/icon/wrapper';
...
 
 export default function CarteProfessionnelleTPSection({
   uniteLegale,
 }: {
   uniteLegale: IUniteLegale;
 }) {
-  const carteProfessionnelleTravauxPublics =
-    useFetchCarteProfessionnelleTP(uniteLegale);
+  const carteProfessionnelleTravauxPublics = useAPIRouteData(
+    'espace-agent/carte-professionnelle-TP',
```
Comment (XavierJp) [inline app/(header-default)/documents/[slug]/_components/carte-professionnelle-TP-section.tsx:16]: The best would be either to call and enum or another typed object with auto completion rather than a string

hooks/fetch/use-API-route-data.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @param slug : slug parameter for the route
+ * @returns {IAPILoading | RouteResponse<T>} - The API loading state or the fetched data
+ */
+export function useAPIRouteData<T extends APIPath>(
+  route: T,
+  slug: string
+): IAPILoading | RouteResponse<T> {
+  const [response, setResponse] = useState<IAPILoading | RouteResponse<T>>(
+    APILoadingFactory()
+  );
```
XavierJp, line 23: Dont we introduce scope checking in front with a `IUnauthorised` ? Or is `DataSection` not ready for this refacto yet ?

hooks/fetch/use-API-route-data.ts
```diff
@@ -0,0 +1,55 @@
...
+          // Errors that are not already logged server side
+          logErrorInSentry(
+            new FailToFetchError({
+              context: {
+                slug: slug,
+                page: '/api/data-fetching/' + route,
+              },
+              cause: e,
+            })
+          );
```
XavierJp, line 47: Nice

XavierJp (commented):
Nice ! Would gladly discuss about it over visio. 

Really like where this is going.

hooks/fetch/use-API-route-data.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Hook to fetch data from internal API
+ * @param route : route to fetch (from {@link APIPath})
+ * @param slug : slug parameter for the route
+ * @returns {IAPILoading | RouteResponse<T>} - The API loading state or the fetched data
+ */
+export function useAPIRouteData<T extends APIPath>(
+  route: T,
+  slug: string
+): IAPILoading | RouteResponse<T> {
+  const [response, setResponse] = useState<IAPILoading | RouteResponse<T>>(
```
XavierJp, line 21: A really nice addition would be to also handle NetworkError here and have the section display a dedicated message "Vérifiez votre connection ou votre VPN etc."

app/api/data-fetching/routes.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const APIHandler = {
+  'espace-agent/carte-professionnelle-TP': getCarteProfessionnelleTravauxPublic,
+  'espace-agent/conformite': NotImplementedHandler,
+  'espace-agent/opqibi': NotImplementedHandler,
+  'espace-agent/qualibat': NotImplementedHandler,
+  'espace-agent/qualifelec': NotImplementedHandler,
+  'espace-agent/rcs-mandataires': NotImplementedHandler,
+  'espace-agent/rne/documents': NotImplementedHandler,
```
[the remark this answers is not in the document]
johangirod: If we do so, it will lead to some network error during the deployment, as we'll change the route of the resource. It's not that much of deal, I think we can be ok with that.

I'm not a big fan of Enum. While they were necessary during the first days of Typescript, we can now have the same type safety power with `as const` without the need to add another specific data structure (and adding one level of indirection over string). 

Their main justification is not to improve typing, but to allow for smoother transition if we decide to rename a route.

app/api/data-fetching/routes.ts
johangirod: I split the two because I'm not sure tree-shaking would work with a single object, and we need scope client side.

app/api/data-fetching/routes.ts
johangirod: Don't have a strong opinion. I think I used `false` because initially the variable was called `APIProtectedRoutes`, and it was making more sense that way. But `null` is also OK.

app/(header-default)/documents/[slug]/_components/carte-professionnelle-TP-section.tsx
Comment (johangirod) [reply app/(header-default)/documents/[slug]/_components/carte-professionnelle-TP-section.tsx:16]: Autocompletion is enabled with `string`, without the need for importing a data object.

XavierJp (commented):
Looks very good. I need a proper laptop to fully test it. I will do this tomorrow

XavierJp (approved):

XavierJp:
@johangirod will merge this tomorrow

PR #1115, opened 2024-06-17
fix: association bilans and middleware coverage

XavierJp:
Two tricky bugs : 

## First one

Caused by DJEPVA mutating siret from string to number. This PR add back open data bilan for association

## second one

Due to next shortcoming. Middleware either apply to every request or you have to add a matcher. Problem : there is no easy matcher for `app`and `pages`. 
As a consequence, visit timestamp is not always set and antibot can get triggered. Not sure this currently happens in prod but this is bug prone.

I refactor a bit the middleware to avoid running the extract siren regex systematically and tried to come up with a more satisfying matcher.


components/finances-section/association/index.tsx
```diff
@@ -24,109 +28,120 @@ const colorCA = constants.chartColors[4];
...
-          <LineChart
-            htmlLegendId={'finance-data-legend'}
-            options={{
-              plugins: {
-                tooltip: {
-                  callbacks: {
-                    label(tooltipItem) {
-                      return formatCurrency(tooltipItem.parsed.y.toString());
+      {(data) =>
+        data?.bilans.length === 0 ? (
```
johangirod, line 51: Je pense que ce genre de cas devrait être géré dans le model, plutôt que dans la vue. Lorsque data.bilans === 0, on retourne un `APINotResponding(404)`.

clients/api-entreprise/association/index.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const mapToDomainObject = (response: IAPIEntrepriseAssociation): unknown => {
+  return {
+    documentsRna: response.data.documents_rna,
+    documentsDac: response.data.etablissements.flatMap((e) => {
+      return { siret: e.siret, ...e.documents_dac };
+    }),
+  };
+};
```
johangirod, line 25: C'est un peu hors sujet sur cette PR non ?

johangirod (approved):

clients/api-entreprise/association/index.ts
XavierJp: Ah mince c'est totalement hors sujet

components/finances-section/association/index.tsx
XavierJp: Oui bonne idée

PR #1170, opened 2024-07-24
feat: responsive and first date

XavierJp:
@johangirod I tried to fix some issues, here are some propositions : 
- rename nouveautés into changelog-notification as it is tied to the changelog logic
- add an initialization date to target visitors that never visited `/historique...`
- responsiveness : I had to fought the header a bit to find a way to make it responsive. Unfortunately I had to remove `<FadeIn />` and the `tooltip` but it turns out that a label on desktop works pretty well (like on DS), and just the Icon on mobile is consistent with the menu icon
- add a early return on `useStorage` when storage is not available

image

image

I really like the icon and the little red dot, catchy but elegant !

### DIFF
--- a/components/header/changelog-notification.module.css
+++ b/components/header/changelog-notification.module.css
@@ -2,7 +2,7 @@
   content: '';
   position: absolute;
   top: 5px;
-  right: 5px;
+  right: 2px;
   width: 10px;
   height: 10px;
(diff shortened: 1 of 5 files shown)


HAEKADI (approved):

HAEKADI:
Very nice!

PR #1232, opened 2024-09-30
fix: suppression insee et recherche entreprise nom commercial

### LINKED ISSUE #1228: Nom commercial
Exemple : 487566606 « SUPPRESSION DU NOM COMMERCIAL ». Probablement un raccourci pris par l'Insee

https://annuaire-entreprises.data.gouv.fr/entreprise/487566606

Todo : 
- ignorer les noms commerciaux `SUPPRESSION DU NOM COMMERCIAL` dans le code du site et de l'API
- informer Anais quand c'est en prod pour qu'elle puisse répondre au ticket

rmonnier9:
- Correction d'un bug
- Zones impactées : `./models/core/unite-legal.ts`.
- Détails :
  - Nettoyage à la volée de la mention "suppression du nom commercial"

closes #1228

### DIFF
--- a/clients/sirene-insee/helpers.ts
+++ b/clients/sirene-insee/helpers.ts
@@ -1,0 +1,10 @@
+/**
+ * When a company removes a commercial name,
+ * INSEE sets its old "denomination usuelle" to "SUPPRESSION DU NOM COMMERCIAL".
+ * This function removes that specific mention.
+ */
+export const formatDenominationUsuelle = (denominationUsuelle: string) => {
+  return denominationUsuelle !== 'SUPPRESSION DU NOM COMMERCIAL'
+    ? denominationUsuelle
+    : '';
+};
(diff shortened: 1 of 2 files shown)


XavierJp (changes requested):
Works pretty well but I would rather suggest a different approach, removing `SUPPRESSION DU NOM COMMERCIAL` directly in the underlying clients rather than i the model

rmonnier9:
The fix is now at the client level.

Is it the correct approach ?
I thought maybe you were referring to an empty field ?

rmonnier9:
I fixed the ‘recherche entreprise’ result, but this might be redundant with the upcoming fix from Hajar.

rmonnier9:
In the example I was using, only “enseigne1Etablissement” was affected, but in case the others can be affected as well, I applied the format function to all three.

XavierJp (approved):

PR #1233, opened 2024-09-30
chore: remove useless fallback on recherche entreprise

XavierJp:
Remove dead code : 
- console 
- old fallback in recherche api

### DIFF
--- a/models/eori-validation.ts
+++ b/models/eori-validation.ts
@@ -20,7 +20,7 @@ export const getEORIValidation = async (
     const siret = verifySiret(eori);
     const data = await clientEORI(siret);
     if (!data) {
-      return { eori, isValid: false };
+      throw new Error('EOS response is empty');
     }
     return data;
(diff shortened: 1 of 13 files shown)


models/eori-validation.ts
```diff
@@ -20,7 +20,7 @@ export const getEORIValidation = async (
     const siret = verifySiret(eori);
     const data = await clientEORI(siret);
     if (!data) {
-      throw new Error('EOS response is empty');
+      return { eori, isValid: false };
```
johangirod, line 23: I don't get why this is necessary. A numero not valid means that the API responded. This ensure that a incorrect response from API will be logged into sentry and will result in a `APINotResponding` error.

johangirod (approved):

models/eori-validation.ts
XavierJp: Makes sense

PR #1239, opened 2024-10-02
fix: good redirect after agent public login

rmonnier9:
- Correctif
- Zones impactées : `./middleware.ts`.
- Détails :
  - The usePathServer hook has been removed.
  - login : The middleware is now responsible for adding the pathFrom param when navigating to the "espace agent" page
  - logout : referer is directly used from the route to determine the pathFrom param

#1220

### DIFF
--- a/components-ui/button-pro-connect/index.tsx
+++ b/components-ui/button-pro-connect/index.tsx
@@ -1,29 +1,32 @@
 'use client';
 
 import { usePathname } from 'next/navigation';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { logConversionEvent } from '#utils/matomo';
 
 type IProps = {
-  useCurrentPathForRediction: boolean;
-  alternatePathForRedirection?: string;
+  shouldRedirectToReferer?: boolean;
   event: string;
 };
 
 const ButtonProConnect: React.FC<IProps> = ({
-  alternatePathForRedirection,
-  useCurrentPathForRediction,
+  shouldRedirectToReferer = false,
   event = 'BTN_DEFAULT',
 }) => {
-  let pathFrom = null;
+  const [referrer, setReferrer] = useState<string | null>(null);
   const currentPath = usePathname();
 
-  if (useCurrentPathForRediction) {
-    pathFrom = currentPath;
-  }
-  if (alternatePathForRedirection) {
-    pathFrom = alternatePathForRedirection;
-  }
+  useEffect(() => {
+    setReferrer(document.referrer);
+  }, []);
+
+  const isFromSite =
+    referrer?.indexOf(
+      process.env.NEXT_PUBLIC_BASE_URL || 'https://annuaire-entreprises'
+    ) === 0;
+
+  const pathFrom =
+    shouldRedirectToReferer && isFromSite ? referrer : currentPath;
 
   return (
     <form action="/api/auth/agent-connect/login" method="get">
--- a/pages/api/auth/agent-connect/logout.ts
+++ b/pages/api/auth/agent-connect/logout.ts
@@ -6,7 +6,7 @@ import withSession from '#utils/session/with-session';
 
 export default withSession(async function logoutRoute(req, res) {
   try {
-    await setPathFrom(req.session, (req?.query?.pathFrom || '') as string);
+    await setPathFrom(req.session, req.headers.referer || '');
     const url = await agentConnectLogoutUrl(req);
     res.redirect(url);
   } catch (e: any) {
--- a/app/(header-default)/lp/agent-public/page.tsx
+++ b/app/(header-default)/lp/agent-public/page.tsx
@@ -1,3 +1,5 @@
+import { IronSession } from 'iron-session';
+import { Metadata } from 'next';
 import { default as ButtonProConnect } from '#components-ui/button-pro-connect';
 import Container from '#components-ui/container';
 import { administrationsMetaData } from '#models/administrations';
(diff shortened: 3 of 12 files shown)


rmonnier9:
I can't figure how to reproduce any other trouble with the button.

XavierJp:
- fix le hook coté server
- nouveau layout du header avec uniquement le search et les logos
- éviter les client component si possible

XavierJp:
I see. Nice job spotting this bug.

XavierJp (approved):
Simple, effective, and remove dead code. Amazing

XavierJp:
Have you tested it extensively ?

XavierJp:
Also, you talked about using the callback from OpenIDConnect? Have you investigated it ?

XavierJp:
@rmonnier9 I love what you did with the referrer. Do you think you can replicate it in order to move this component from a client component using `usePathname` into a server component ? 
https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/6977405015ec45cf116ba9a61b5f646d997f1c0b/components-ui/button-pro-connect/index.tsx#L1

rmonnier9:
> Also, you talked about using the callback from OpenIDConnect? Have you investigated it ?

Yes, but in fact, it would have been a way to replace the data we save in session cookie between the initial call and the callback.

And actually, I think it was something particular in an other API I was using.

rmonnier9:
> @rmonnier9 I love what you did with the referrer. Do you think you can replicate it in order to move this component from a client component using `usePathname` into a server component ?
> 
> https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/6977405015ec45cf116ba9a61b5f646d997f1c0b/components-ui/button-pro-connect/index.tsx#L1
> 
> There is a challenge :
> 
> * button is either called from here https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/6977405015ec45cf116ba9a61b5f646d997f1c0b/components/espace-agent-components/agent-wall/index.tsx#L1
> * or from the landing page
> 
> So:
> 
> * When calling from `AgentWall` you want the pathFrom to be the current page
> * When callfing from LP you want pathFrom to be the referrer

Regarding the browser history, the list of previous pages is not accessible due to security reasons.

As an alternative, I found a solution: calling an intermediate API route that redirects to the correct page while adding the pathFrom parameter.

XavierJp:
Becoming really nice. I think we can use an even simpler implem : `pages/api/auth/agent-connect/redirect.ts`  with the following `ButtonProConnect` :

```
const ButtonProConnect: React.FC<IProps> = ({
  shouldRedirectToReferer=false
  event = 'BTN_DEFAULT',
}) => {

	const currentPath = usePathname();
	// it seems we can access the referrer from client https://developer.mozilla.org/fr/docs/Web/API/Document/referrer
	const referrer = document?.referrer; 
 	cont pathFrom = shouldRedirectToReferer ? referrer : currentPath;
 	
 	return (
    <form action="/api/auth/agent-connect/login" method="get">
      {pathFrom && (
        <input readOnly hidden aria-hidden name="pathFrom" value={pathFrom} />
      )}
	...

```

rmonnier9:
> /lp/agent-public?pa

Brilliant. It will be so simple with this !

rmonnier9:
> Becoming really nice. I think we can use an even simpler implem : `pages/api/auth/agent-connect/redirect.ts` with the following `ButtonProConnect` :
> 
> ```
> const ButtonProConnect: React.FC<IProps> = ({
>   shouldRedirectToReferer=false
>   event = 'BTN_DEFAULT',
> }) => {
> 
> 	const currentPath = usePathname();
> 	// it seems we can access the referrer from client https://developer.mozilla.org/fr/docs/Web/API/Document/referrer
> 	const referrer = document?.referrer; 
>  	cont pathFrom = shouldRedirectToReferer ? referrer : currentPath;
>  	
>  	return (
>     <form action="/api/auth/agent-connect/login" method="get">
>       {pathFrom && (
>         <input readOnly hidden aria-hidden name="pathFrom" value={pathFrom} />
>       )}
> 	...
> ```
> 
> With this implem we completly get rid of most `pathFrom` in url params. Ithink we should still offer the possibility to create a `/lp/agent-public?pathFrom=XXX` but it is no longer necessary to use it
> 
> Also, referrer has to be compared with your new BASE_URL to avoid redirect to a different origin

Brilliant ! I will be really really simple with this.

components-ui/button-pro-connect/index.tsx
```diff
@@ -1,29 +1,29 @@
...
 
-  if (useCurrentPathForRediction) {
-    pathFrom = currentPath;
-  }
-  if (alternatePathForRedirection) {
-    pathFrom = alternatePathForRedirection;
-  }
+  useEffect(() => {
+    setReferrer(document.referrer);
+  }, []);
```
XavierJp, line 21: Improvement idea : create a useReferer hook as we might need it somewhere else

components-ui/button-pro-connect/index.tsx
```diff
@@ -1,29 +1,29 @@
...
+
+  const isFromSite = referrer?.indexOf('https://annuaire-entreprises') === 0;
```
XavierJp, line 23: Improvement idea : use base url or a loose regex to make this work also in staging

pages/api/auth/agent-connect/logout.ts
```diff
@@ -6,7 +6,7 @@ import withSession from '#utils/session/with-session';
 
 export default withSession(async function logoutRoute(req, res) {
   try {
-    await setPathFrom(req.session, (req?.query?.pathFrom || '') as string);
+    await setPathFrom(req.session, req.headers.referer || '');
```
XavierJp, line 9: Haha that is so plain simple and smart. Genius

XavierJp (approved):
Sounds good to me ! You can merge it

XavierJp:
Test e2e fail looks like noise. The tests checks wether footer has any broken links. So we cannot moke this part. But the error is not relevant here. Looks like a temporary downtime from remote website

PR #1249, opened 2024-10-04
style: improve agent banner

XavierJp:
reference #1247 

- Does not add a link to tchap but improve espace agent top banner
- Clean unused layout


XavierJp:
Yet another PR that removes code ! ! !

PR #1257, opened 2024-10-08
fix: incorrect 500 in static pages

XavierJp:
In page router we use to throw exception and then handle it later.

With app router we must call `notFound()` instead


app/(header-default)/administration/[slug]/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { administrationsMetaData } from '#models/administrations';
+import { EAdministration } from '#models/administrations/EAdministration';
+import { getFaqArticlesByTag } from '#models/article/faq';
+import { AppRouterProps } from '#utils/server-side-helper/app/extract-params';
+
+const cachedGetAdministrations = cache((slug: string) => {
+  const slugs = slug.split('_');
+
+  const administrations = Object.values(administrationsMetaData).filter(
+    //@ts-ignore
```
Comment (rmonnier9) [inline app/(header-default)/administration/[slug]/page.tsx:15]:
```suggestion
```

rmonnier9 (approved):

PR #1266, opened 2024-10-11
chore: migrate api routes from pages to app

### LINKED ISSUE #1267: Migrate auth api from pages to app

rmonnier9:
- Migration
- Zones impactées : `api/auth/agent-connect`, `api/auth/france-connect`.
- Détails :
  - Migration sur App Router

Closes #1267 

Next thing : connexion pages migration in another PR (this one is required before merging agent habilitation PR)


XavierJp:
Could not make it work, both AgentConnect and FranceConnect does not work in local. FranceConnect might be unrelated to your changes though

rmonnier9:
> Could not make it work, both AgentConnect and FranceConnect does not work in local. FranceConnect might be unrelated to your changes though

I fixed it.

When using NextResponse.redirect, full url must be used.
There is a builtin function in nextjs that allows us to use path (redirect from next/navigation) but it throws errors in its normal behavior (to stop rendering a server component for example). This behaviour messes up with our try catch.
I created an utils redirectTo that creates the url.

image

app/api/auth/utils.ts
XavierJp, line 1: I think this should be moved in `/utils/server-side-helper/app/redirect-to.ts`

XavierJp (changes requested):

PR #1271, opened 2024-10-14
Bump package - possible memory leaks

XavierJp:
Reverts annuaire-entreprises-data-gouv-fr/site#1270


XavierJp:
relevevant packages have been bumped with #1290

PR #1274, opened 2024-10-15
Fix france connect

rmonnier9:
- Correction d'un bug
- Zones impactées : `/api/auth`.
- Détails :
  - change redirectTo to getURL

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -81,8 +81,7 @@ export async function middleware(request: NextRequest) {
     // store redirection status in custom header as referrer seems missing from headers in RSC
     // isRedirected = params is present + previous page is coming from site
     const referer = requestHeaders.get('referer') || '';
-    const baseURL =
-      process.env.NEXT_PUBLIC_BASE_URL || 'https://annuaire-entreprises';
+    const baseURL = process.env.NEXT_PUBLIC_BASE_URL || 'http://localhost:3000';
     const isFromSite = referer.indexOf(baseURL) === 0;
 
     const isRedirected = paramIsPresent && isFromSite ? '1' : '0';
--- a/middleware.ts
+++ b/middleware.ts
@@ -81,7 +81,9 @@ export async function middleware(request: NextRequest) {
     // store redirection status in custom header as referrer seems missing from headers in RSC
     // isRedirected = params is present + previous page is coming from site
     const referer = requestHeaders.get('referer') || '';
-    const baseURL = process.env.NEXT_PUBLIC_BASE_URL || 'http://localhost:3000';
+    const baseURL =
+      process.env.NEXT_PUBLIC_BASE_URL ||
+      'https://annuaire-entreprises.data.gouv.fr';
     const isFromSite = referer.indexOf(baseURL) === 0;
 
     const isRedirected = paramIsPresent && isFromSite ? '1' : '0';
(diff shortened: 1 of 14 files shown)


middleware.ts
```diff
@@ -81,8 +81,7 @@ export async function middleware(request: NextRequest) {
     // store redirection status in custom header as referrer seems missing from headers in RSC
     // isRedirected = params is present + previous page is coming from site
     const referer = requestHeaders.get('referer') || '';
-    const baseURL =
-      process.env.NEXT_PUBLIC_BASE_URL || 'https://annuaire-entreprises';
+    const baseURL = process.env.NEXT_PUBLIC_BASE_URL || 'http://localhost:3000';
```
XavierJp, line 84: Sounds bug prone, default should be prod url rather than local. If one forgot to declare `NEXT_PUBLIC_BASE_URL` it is OK if local fails, but not if prod fails because of it. I prefer 

`const baseURL = process.env.NEXT_PUBLIC_BASE_URL || 'https://annuaire-entreprises.data.gouv.fr';`

utils/server-side-helper/app/get-url.ts
```diff
@@ -0,0 +1,7 @@
+export const getURL = (path: string = '') => {
+  const baseURL = process.env.NEXT_PUBLIC_BASE_URL || 'http://localhost:3000';
```
XavierJp, line 2: Idem for default url. You could even declare a constant `prodBaseUrl` here https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/08f8511c0e90b82ed4e191bb81419483d7ca423f/models/constants.ts#L1

utils/server-side-helper/app/get-url.ts
```diff
@@ -0,0 +1,7 @@
+export const getURL = (path: string = '') => {
```
XavierJp, line 1: getUrl is very generic. A more self explaining function name and a few comments will improve readability 

like `getAbsoluteSiteUrl()`

XavierJp (approved):
I like it better than `redirectTo` !

utils/server-side-helper/app/get-url.ts
XavierJp: Actually not sure`prodBaseUrl` is a good idea

middleware.ts
XavierJp: Actually not sure`prodBaseUrl` is a good idea

rmonnier9: I did it this way because most of the time in other repos it is written like that.

rmonnier9: I renamed to getAbsoluteSiteUrl, good idea !

utils/server-side-helper/app/get-url.ts
rmonnier9: I renamed to getAbsoluteSiteUrl, good idea !

rmonnier9:
@MKCG required before merge : tell us when the new env vars are deployed !

middleware.ts
rmonnier9: I'm okay with the two versions

XavierJp: I see but the way we use it, it means that if we forget it while provisionning a new machine, it will break prod. 

I would consider it a good practise to use loacalhost if for instance we would  use it to log errors or stats. Better log them as localhost by default in order not to polluate our analytics

XavierJp: I prefer to have a working default otherwise we have to add tests to our provisionning scripts

XavierJp (approved):

XavierJp:
Thx ! Ready for merge

XavierJp:
@rmonnier9 I suggest we merge this as default variable ensure it will work in prod and currently FranceConnect does not work and we have to hotfix it

PR #1284, opened 2024-10-18
fix: link to bodacc

### LINKED ISSUE #86: Erreur sur un lien dans l'onglet annonces
Bonjour, je viens de tomber sur un lien qui ne marche pas. Ca semble reproductible sur toutes les entreprises.

Exemple : 
1) aller sur https://annuaire-entreprises.data.gouv.fr/annonces/306138900
2) cliquer sur le lien ` la page de cette entreprise`
3) le lien https://www.bodacc.fr/annonce/liste//306138900 a un double `/`

J'ai été tenté par une PR, mais je ne maitrise pas la stack utilisé, et j'ai abandonné la recherche de trouver le template correspondant :)

rmonnier9:
- Correction d'un bug.
- Détails :
  - Fix du lien
  - Refacto every replace with functions

Closes #86

### DIFF
--- a/app/(header-default)/annonces/[slug]/_components/bodacc.tsx
+++ b/app/(header-default)/annonces/[slug]/_components/bodacc.tsx
@@ -68,7 +68,7 @@ const AnnoncesBodacc: React.FC<{
                 consulter{' '}
                 <UniteLegalePageLink
                   uniteLegale={uniteLegale}
-                  href={`${routes.bodacc.site.recherche}/${uniteLegale.siren}`}
+                  href={routes.bodacc.site.rechercheBySiren(uniteLegale.siren)}
                   siteName="le site du BODACC"
                 />
                 &nbsp;:
--- a/clients/api-data-subvention/index.ts
+++ b/clients/api-data-subvention/index.ts
@@ -12,7 +12,7 @@ import { httpGet } from '#utils/network';
 export const clientDataSubvention = async (
   siren: Siren
 ): Promise<ISubventions> => {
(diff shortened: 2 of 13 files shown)


app/(header-default)/annonces/[slug]/_components/bodacc.tsx
```diff
@@ -68,7 +68,10 @@ const AnnoncesBodacc: React.FC<{
                 consulter{' '}
                 <UniteLegalePageLink
                   uniteLegale={uniteLegale}
-                  href={`${routes.bodacc.site.recherche}/${uniteLegale.siren}`}
+                  href={routes.bodacc.site.rechercheBySiren.replace(
+                    '{siren}',
+                    uniteLegale.siren
+                  )}
```
Comment (XavierJp) [inline app/(header-default)/annonces/[slug]/_components/bodacc.tsx:74]: Would prefer an url without the replace when possible

XavierJp (changes requested):

XavierJp (commented):
Sounds good to me. Only tiny issue I see is that you mutate route.ts signature. Now it can either be a function or a string. Why not only have string ?

XavierJp (approved):

rmonnier9:
> Sounds good to me. Only tiny issue I see is that you mutate route.ts signature. Now it can either be a function or a string. Why not only have string ?

If we want to keep only strings in routes.ts, I don’t see another way than using placeholders and replacing them with .replace(). We could create a utility function to handle this for us, but it might make it less transparent what’s happening under the hood.

rmonnier9:
I'm actually ok with both options

XavierJp:
I like your current solution

PR #1311, opened 2024-10-24
feat: add sentry log for grist errors

rmonnier9:
- Amélioration technique.
- Détails :
  - Add Sentry Grist


XavierJp (approved):
Nice. Might create duplicated log but should not happen often 

For instance this. will log two errors
https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/e14893406cdf6bfa58f79ae79ddfc0e9d3f6cab7/clients/authentication/super-agent-scopes/index.ts#L11

PR #1318, opened 2024-10-28
Bump next 15

rmonnier9:
- Amélioration technique. 
- Détails :
  - Mise à niveau vers Next 15


rmonnier9:
Actually, I just had to remove :
const ChangelogNotificationWithoutSSR = dynamic(
  () => import('../changelog-notification'),
  {
    ssr: false,
  }
);

and import a regular component to make it works without 'use client'.

utils/server-side-helper/app/is-bot.ts
XavierJp, line 1: Seems like it is not used anymore. can you remove the file ?

XavierJp (approved):
I fear the changelog notif might trigger an hydration issue but worth trying as is

XavierJp (approved):

PR #1319, opened 2024-10-29
fix: add generateBuildId in next.config.js

rmonnier9:
- Correction d'un bug.
- Zones impactées : `next.config.js`.
- Détails :
  - Could fix some chunk load errors

Need to be discussed.

Details from the doc of `next-build-id` :

> Why?
> If you're running multiple instances of your app sitting behind a load balancer without session affinity (and you're building your app directly on each production server instead of pre-packaging it), a tool like this is necessary to avoid Next.js errors like ["invalid build file hash"](https://github.com/zeit/next.js/blob/52ccc14059673508803f96ef1c74eecdf27fe096/server/index.js#L444), which happens when the same client (browser code) talks to multiple server backends (Node server) that have different build ids.
> 
> The build id used by your app is stored on the file system in a BUILD_ID text file in your build directory, which is .next by default.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,5 +1,6 @@
 const { withSentryConfig } = require('@sentry/nextjs');
 const redirects = require('./redirects.json');
+const nextBuildId = require('next-build-id');
 
 const WITH_SENTRY =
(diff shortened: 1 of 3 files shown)


XavierJp (approved):
This is lit!

PR #1326, opened 2024-10-31
chore: remove useless error file

rmonnier9:
- Refacto
- Détails :
  - Remove useless error file


app/(header-default)/erreur/introuvable/[slug]/page.tsx
```diff
@@ -1,31 +1,45 @@
...
+  const isValid = isLuhnValid(slug);
+
+  return {
+    title: isValid
+      ? 'Numéro d’identification introuvable'
+      : 'Numéro d’identification invalide',
+    robots: { noindex: true },
+  };
+}
+export const notFoundRegexpForStatusCode = /^\/erreur\/introuvable\/(\d{9})$/;
```
Comment (XavierJp) [inline app/(header-default)/erreur/introuvable/[slug]/page.tsx:27]: What does it do ?

app/(header-default)/erreur/trop-de-demandes/page.tsx
```diff
@@ -1,13 +1,16 @@
...
-import { NextPageWithLayout } from 'pages/_app';
 
-const TooManyRequest: NextPageWithLayout = () => {
+export async function generateMetadata() {
+  return {
+    title: 'Trop de requête de votre part',
+    robots: { noindex: true },
+  };
+}
+export const tooManyRequestsPathForStatusCode = '/erreur/trop-de-demandes';
```
XavierJp, line 10: I think I start to undestand, but these regex need a comment, maybe a link to the documentation

middleware.ts
```diff
@@ -89,11 +94,26 @@ export async function middleware(request: NextRequest) {
...
+  } else if (
+    request.nextUrl.pathname.match(notFoundRegexpForStatusCode) ||
+    request.nextUrl.pathname === page404PathForStatusCode
+  ) {
+    status = 404;
+  } else if (request.nextUrl.pathname === tooManyRequestsPathForStatusCode) {
+    status = 429;
+  } else if (request.nextUrl.pathname === page500PathForStatusCode) {
+    status = 500;
+  }
```
XavierJp, line 109: This looks a bit complicated and bug prone compared to the old way

XavierJp (changes requested):
We should talk about this

rmonnier9:
It now only removes a useless error file.

PR #1370, opened 2024-11-13
feat: add header no-sniff

rmonnier9:
- Amélioration technique.
- Zones impactées : `next.config.ts`.
- Détails :
  - Ajout du header nosniff
 
https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/X-Content-Type-Options

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -39,6 +39,19 @@ const nextConfig: NextConfig = {
   async redirects() {
     return redirects;
   },
+  async headers() {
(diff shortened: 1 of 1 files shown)


XavierJp (commented):
I am not sure about this (I mean we should definitely have this header, but should we set it here or at nginx level as part of a general `X-Content-Type-Options` improvement ?)
requesting @MKCG review on this

PR #1373, opened 2024-11-15
fix: use crypto for random bytes

rmonnier9:
- Amélioration technique.
- Détails :
    - Lié à https://github.com/annuaire-entreprises-data-gouv-fr/site/security/code-scanning/2

### DIFF
--- a/utils/helpers/random-id.ts
+++ b/utils/helpers/random-id.ts
@@ -1,1 +1,5 @@
-export const randomId = () => Math.random().toString(16).substring(7);
+export const randomId = () => {
+  const crypto =
+    typeof window === 'undefined' ? require('crypto') : window.crypto;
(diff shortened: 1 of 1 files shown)


XavierJp (approved):

XavierJp:
This is. nice beacause it is good practise, but lets face it randomId was never used in a security-related context
`Using a cryptographically weak pseudo-random number generator to generate a security-sensitive value, such as a password, makes it easier for an attacker to predict the value.`

rmonnier9:
> This is. nice beacause it is good practise, but lets face it randomId was never used in a security-related context `Using a cryptographically weak pseudo-random number generator to generate a security-sensitive value, such as a password, makes it easier for an attacker to predict the value.`

True, but at least we don't get anymore the github security notification about it :)

PR #1409, opened 2024-11-26
fix: 400 in IG and warn in local

XavierJp:
- fix issue with missing partial dates
- fix issue with roles in lower case
- fix EI issue


clients/api-proxy/rne/index.ts
```diff
@@ -103,7 +104,7 @@ const mapToDomainObject = ({
       siren: personneMorale.siren,
       denomination: personneMorale.denomination,
       natureJuridique: personneMorale.natureJuridique,
-      role: personneMorale.role,
+      role: (personneMorale.role || '').toUpperCase(),
```
rmonnier9, line 107: In the case where it is :
INPI: Directeur Général
IG: DIRECTEUR GENERAL
Maybe it won't work ?

We could create an util for roles that also remove special characters :
export const formatRole = (role: string) => {
  if (!role) {
    return '';
  }
  return removeSpecialChars(role).toUpperCase();
};

rmonnier9: Also, should we apply this rule to the rechercheEntreprise client ?

rmonnier9 (approved):

PR #1412, opened 2024-11-27
chore: add security headers on site side rather than infra

XavierJp:
As suggested by @ad2ien, writing security headers on site side allow dev to have much more control on POST routes and cross domain calls.

The alternative is to write them at nginx level and have every headers at the same place. @MKCG @rmonnier9 @ad2ien what do you think ? 

closes https://github.com/orgs/annuaire-entreprises-data-gouv-fr/projects/1/views/1?pane=issue&itemId=84349585&issue=annuaire-entreprises-data-gouv-fr%7Cinfrastructure%7C234


rmonnier9 (approved):

ad2ien (approved):
:+1:

PR #1419, opened 2024-11-29
fix: remove hydration errors

rmonnier9:
- Amélioration technique. 
- Détails :
  - Création d'un composant ClientOnly destiné à éviter les erreurs d'hydratation
  - Amélioration du typage de map
  - Utilisation de ClientOnly sur le multi select et sur la bannière NPS

Avant : 
image

Après :
image

### DIFF
--- a/components-ui/select/multi-select.tsx
+++ b/components-ui/select/multi-select.tsx
@@ -1,3 +1,4 @@
+import ClientOnly from '#components/client-only';
 import ReactSelect, { MultiValue, Props } from 'react-select';
 import { ISelectOptions } from './type';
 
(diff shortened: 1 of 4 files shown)


XavierJp (approved):
This is a smart move. I love the `<ClientOnly />`

Could you maybe please add a few comments in components/client-only/index.tsx on the why, how to use and the differences between `<ClientOnly/>` and `use client`

rmonnier9:
> This is a smart move. I love the `<ClientOnly />`
> 
> Could you maybe please add a few comments in components/client-only/index.tsx on the why, how to use and the differences between `<ClientOnly/>` and `use client`

I've added :
// Even with 'use client', Next.js pre-renders components on the server.
// This wrapper delays rendering until the component has mounted on the client.

PR #1424, opened 2024-12-02
chore: add keys and remove some ts-ignore

rmonnier9:
- Changement mineur.
- Détails :
  - Add some keys
  - Remove some ts-ignore


XavierJp (approved):
Very nice

PR #1464, opened 2024-12-16
chore: remove stub for ods

rmonnier9:
- Amélioration technique.
- Détails :
  - Remove stub for ods


cypress/e2e/bilans-financiers.cy.js
```diff
@@ -4,23 +4,23 @@ describe('Bilans financiers', () => {
...
 
-    cy.contains('Résultat net');
-    cy.contains('2 Mds €');
+  //   cy.contains('Résultat net');
+  //   cy.contains('2 Mds €');
 
-    cy.contains('Source : MEF');
-  });
+  //   cy.contains('Source : MEF');
+  // });
```
XavierJp, line 23: Either delete or uncomment ?

XavierJp (approved):

cypress/e2e/bilans-financiers.cy.js
rmonnier9: When I visit either :
https://annuaire-entreprises.data.gouv.fr/donnees-financieres/552032534'
or
https://annuaire-entreprises.data.gouv.fr/donnees-financieres/487444697

image

There is nothing.
The test was working before removing stub because of the data inside the stub.

I commented it until I rework with the new data sources.
What do you think about that ?

XavierJp: Source is currently down. Which is annoying as you cannot create mock ^^

PR #1484, opened 2024-12-23
fix(data subvention): try to use id rna

rmonnier9:
- Correction d'un bug.
- Détails :
  - Récupération via l'id rna (ou siret si id rna est nulle)
  - Puis fallback sur siret
  - Correctif d'affichage (ajout de marge)

Avant : 
image

Après :
image


models/subventions/association/index.ts
```diff
@@ -26,24 +26,34 @@ export const getSubventionsAssociationFromSlug = async (
...
+          return APINotRespondingFactory(EAdministration.DJEPVA, 404);
+        }
+        logErrorInSentry(
+          new FetchRessourceException({
+            ressource: 'DataSubvention',
+            cause: e,
+            context: {
+              siren,
+            },
+            administration: EAdministration.DJEPVA,
```
XavierJp, line 52: Sounds great, I see two (minor) improvements : 
- could be interesting to have temporary logs in order to know if this work. A `logInfoInSentry` that says fallback worked for association XXX should be enough 
- current implem will call twice with siren if no IdAssociation. Maybe prefer a first call with siren and then, when it fails : 

```
...
catch(e) {
	if (e instanceof HttpNotFound) {
	          return APINotRespondingFactory(EAdministration.DJEPVA, 404);
        }
	if (idAssociation) {
		try {
			return await clientApiDataSubvention(idAssociation);
		} catch...
	} else {
		return APINotRespondingFactory(EAdministration.DJEPVA, 500);
	}
}
```

XavierJp (approved):

XavierJp (approved):

PR #1495, opened 2024-12-29
feat(dgfip liasses fiscales): add dgfip liasses fiscales data

### LINKED ISSUE #1492: Nouvelle donnée : DGFIP Bilans

rmonnier9:
- Fonctionnalité
- Détails :
  - Ajout de la source de données DGFIP Liasses fiscales

Closes #1492


rmonnier9:
> We should add a
> 
> Plus there are some weird double refresh issues, is not displayed when it should

What should we add ?

I'll check the double refresh issue

PR #1535, opened 2025-01-14
Update data store with auto refresh strategy

XavierJp:


rmonnier9 (approved):

clients/data-store.ts
```diff
@@ -0,0 +1,53 @@
+import { HttpServerError } from '#clients/exceptions';
+import { FetchRessourceException } from '#models/exceptions';
+import { logWarningInSentry } from '#utils/sentry';
+
+/**
+ * Generic client that cache an API response in memory
```
rmonnier9, line 6:
```suggestion
 * Generic client that caches an API response in memory
```

rmonnier9 (approved):

PR #1542, opened 2025-01-16
fix(external-api-test): sign commits 2

rmonnier9:
Deuxième tentative pour signer les commits faits par le bot.

Documentation de `peter-evans/create-pull-request@v7` : "In this example the token input is not supplied, so the action will use the repository's default GITHUB_TOKEN. This will sign commits as github-actions[bot]."

image


XavierJp (approved):
Let's try it !

PR #1553, opened 2025-01-17
feat: update conformité

XavierJp:
- update labels
- add use case to agents logs


components/section-with-use-case/index.tsx
```diff
@@ -19,6 +19,15 @@ type WrappedSectionProps = {
   isProtected: boolean;
 };
 
+const DefaultIntroContent = () => (
+  <p>
+    Les agents publics peuvent demander accéder à cette donnée uniquement dans
```
MKCG, line 24:
```suggestion
    Les agents publics peuvent demander à accéder à cette donnée uniquement dans
```

MKCG (approved):

HAEKADI (approved):

PR #1586, opened 2025-02-06
log data sub in kibana

XavierJp:


HAEKADI (approved):
What are you logging exactly?

PR #1589, opened 2025-02-10
turn the getData promise into a singleton

XavierJp:
## Problem 

DataStore is a singleton that fetch and store data. It has a TTL and tries to refresh it every time the TTL has expire.

BUT with current implementation : 
- on first call ever to the datastore, it triggers `getData` an async http call that fetch data, and setup a setTimeout for refresh in TTL millisecondes
- if DataStore get another request, it triggers another time getData and the setTimeout for refresh
- and so on. It can keep triggering requests as long as the first `getData` has not been resolved

This can trigger several simultaneous setTimeout loops, triggered more or less at the same time. It can create burst calls to a remote API and thus get us some sweet 429 😬

## Solution

Let's turn the getData into singleton that every requests await when it is not `null`.


XavierJp:
Not sure about this pattern, memory leaks and so on. Would need some expert eye on this

HAEKADI (approved):

utils/data-store/index.ts
```diff
@@ -0,0 +1,77 @@
...
+import { FetchRessourceException } from '#models/exceptions';
+import { logWarningInSentry } from '#utils/sentry';
+
+/**
+ * Caches a promise response in memory
+ *
+ * When gave a TTR, it will attempt to refresh data every TTR milliseconds.
+ * As long as it as a data in store it will ignore any failed attempt to refresh and still return outdated cached data
+ */
+export class DataStore<T> {
```
MKCG, line 11: Ce n'est pas un singleton : il peut y avoir plusieurs instances de DataStore qui coexistent.

Néanmoins chaque type d'instance (storeName) du DataStore est actuellement instantié une seule fois du fait de leur encapsulation au sein de variables globales communes à l'ensemble des requêtes (superAgentsList, clientIdccMetadata, etc).

MKCG (approved):

utils/data-store/index.ts
XavierJp: Yes en effet. Après par singleton, je faisais plutot référence a l'utilisation de `this.onGoingRefresh` pour dire que la requête de refresh doit forcément être unique

XavierJp: Mais je suis d'accord, le terme est abusif

PR #1651, opened 2025-03-14
fix: end-to-end tests for footer links

### LINKED ISSUE #1639: [BUG] Mettre les tests end2end de 404 de liens externes dans un cron job hebdomadaire
## Sujet du problème

Décrivez votre problème ici.

## Comment le reproduire ?

Étapes à suivre pour faire apparaître le bug :
1. Aller à `…`
2. Cliquer sur `…`
3. Naviguer jusqu’à `…`
4. Constater le bug

## Comportement attendu

Décrivez-nous ce à quoi vous vous attendiez.

## Screenshots

Le cas échéant, ajoutez des captures d'écran pour aider à expliquer votre problème.

rmonnier9:
- Amélioration technique.
- Détails :
  - Remove e2e tests to external links as they fail randomly
  - Add dedicated worflow for external links
 
Closes #1639

### DIFF
--- a/cypress/e2e/fonctionnalites/static-pages.cy.js
+++ b/cypress/e2e/fonctionnalites/static-pages.cy.js
@@ -24,7 +24,6 @@ pages.forEach((page) => {
 
     it('Logo button works', () => {
       cy.visit(page);
-
       cy.get('.fr-header__logo > a').click();
 
       // Should be on a new URL which includes '/commands/actions'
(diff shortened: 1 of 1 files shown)


XavierJp (approved):
Sounds good ! 

A further improvement could be : create dedicated cypress.config files that specify files to include/exclude.

Also would it be ok to test production directly ? Might be a bit flaky if Baleen randomly block cypress though.

cypress/e2e/fonctionnalites/external-links.cy.js
```diff
@@ -0,0 +1,19 @@
...
+      const currentOrigin = window.location.origin;
+
+      // Only test internal links (same origin)
+      // Skip departements/index.html as it only exists in production and staging
+      if (
+        !href.startsWith(currentOrigin) &&
+        href.indexOf('departements/index.html') === -1
+      ) {
+        cy.request(href);
+      }
```
XavierJp, line 16: Not sure about the comment.

rmonnier9: indeed

XavierJp (commented):
It looks like nothing is running èxternal-link` anymore isnt it ?

rmonnier9:
> It looks like nothing is running `external-link` anymore isnt it ?

Yes I removed it because I don't know if we should really test it every week - at the moment it will fail each time.
The problem is the workflow gets blocked by external sites.
Maybe we should only run it locally when we want to test ?

XavierJp:
Than why not delete it entirely ? Looks a lot like dead code to me :)

rmonnier9:
> Then why not delete it entirely ? Looks a lot like dead code to me :)

I removed it :)

PR #1654, opened 2025-03-14
fix: improve failed workflow message

rmonnier9:
- Amélioration technique. 
- Détails :
  - improve failed workflow formatting


XavierJp (approved):
Not sure if this is necessary, as SEO workflow already notifies us on success. Only failure seems needed as if you fail to trigger, SEO workflow might also fail to notify us of failure

PR #1708, opened 2025-04-07
fix: correct spelling of "allégée" to "allégé"

rmonnier9:


XavierJp (approved):

XavierJp:
I fear there might be other occurences of allégéE

PR #1561, opened 2025-01-20
feat: dgfip and inpi bce finance data comparison

rmonnier9:
- Fonctionnalité
- Détails :
  - Comparaison entre INPI/BCE et DGFiP pour les données financières

#1512 

Sélecteur de cas d'usage sans données open data :
image

Sélecteur de cas d'usage avec données open data :
image

Données différente :
image

PR #1709, opened 2025-04-07
feat: add use case select for Bilans Banque de France section

### LINKED ISSUE #1707: [Banque de France] Use case select

rmonnier9:
- Changement mineur.
- Détails :
  - Use case select for Banque de France bilans

Closes #1707 

image


XavierJp (approved):
I think it might be better to first merge #1561

Especially in 1561 I added an explanation for the companies younger than three years

PR #1798, opened 2025-05-20
feat: manager users in team using d-roles

rmonnier9:
- Amélioration technique. | Correction d'un bug. | Changement mineur.
- Zones impactées : `chemin/vers/le/fichier/contenant/les/variables/impactées`.
- Détails :
  - Ajout d'utilisateur dans une team
  - Suppression d'utilisateur dans une team

Closees #1786


models/authentication/agent/agent-connected/index.tsx
```diff
@@ -92,7 +93,28 @@ export class AgentConnected {
...
+    // Get all groups this agent belongs to from D-Roles API
+    const drolesGroups = await getGroupsByEmail(this.email);
+
+    // Each group has a space-separated string of scopes
+    // Extract all scopes from all groups and flatten into single array
+    const superAgentScopesWithDuplicates = drolesGroups
+      .map((group) => {
+        return group.scopes.split(' ') as IAgentScope[];
+      })
+      .flat();
```
XavierJp, line 105: All this would be nice into a dedicated model. Maybe a groups model, that is a wrapper around d-roles client

models/authentication/agent/agent-connected/index.tsx
```diff
@@ -92,7 +93,28 @@ export class AgentConnected {
...
+    const superAgentScopesFromS3 = await superAgentsList.getScopeForAgent(
+      this.email
+    );
+
+    // Combine scopes from both D-Roles groups and S3 storage
+    // Remove any duplicate scopes using Set
+    const superAgentScopes = [
+      ...new Set(superAgentScopesWithDuplicates),
+      ...new Set(superAgentScopesFromS3),
+    ];
```
XavierJp, line 117: Merging a list of scopes is also done line 102. Maybe create a helper function in `../scopes`that is used to merge scope and unit tested as this is core business logic

XavierJp (approved):
Very nice. I think this would benefit from some refactoring to match existing architecture and code pattern.
I think you need to decide what to do when a user belong to several groups. 

Will he be able to use both groups or will he have to choose which group he is connected with ?

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    tokenExpiryTime: number;
+  } | null;
+
+  constructor(
+    private client_id: string | undefined,
+    private client_secret: string | undefined
+  ) {
+    if (
+      (!this.client_id || !this.client_secret) &&
+      process.env.NODE_ENV === 'production'
```
XavierJp, line 31: Condition is a bit hard to read. I think the test on client_id and client_secret is enough

XavierJp: I believe in staging we do have `process.env.NODE_ENV === 'production'`

I am not sure about this whole condition and how it will work in production

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,156 @@
...
+          client_id: this.client_id || '',
+          client_secret: this.client_secret || '',
+          grant_type: 'client_credentials',
+        }).toString(),
+      });
+
+      this._token = {
+        data,
+        tokenExpiryTime: new Date().getTime() + 240 * 1000, // 4 minutes default expiry
+      };
```
XavierJp, line 57: Why not use the token expiry date returned by d-roles ?

It should contain : 
```
{
    "access_token",
    "token_type",
    "expires_in"
}
```

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    return httpGet<T>(url, {
+      timeout: constants.timeout.M,
+      ...config,
+      headers: {
+        ...config.headers,
+        Authorization: `Bearer ${token.data.access_token}`,
+      },
+    });
+  };
```
XavierJp, line 100: What happens when token is expired ? In InseeClient, I believe that token get reset to `null` thus ensuring you will re-run `getToken`

Also I think splitting client from the rest could improve readibility : 
```
client.tsx // auth + basic POST and GET wrapper
interface.tsx // types
index.tsx // all the routes wrapper
```

I advocate for this code splitting as it leave more room for core routes logic : 
- in case you want to do a mapToDomainObject, you can do it in `index.tsx`
- in case you want to have a file per group of routes you could even have `groups.ts, users.ts, roles.ts`

What you have written works great, as long as you dont use more routes or run heavier manipulation that just a basic `await`

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    groupId: number,
+    email: string,
+    roleId: number
+  ): Promise<null> => {
+    const user = await this.getUserByEmail(email);
+    const route = routes.dRoles.groups.addUserToGroup(groupId, user.id, roleId);
+    return await this.get<null>(route, {
+      method: 'PUT',
+      data: { email },
+    });
```
XavierJp, line 136: Calling a `.get` with a `PUT` sound odd.
Maybe rename `get` into `client` and use `httpClient` rather than `httpGet`

clients/routes.ts
```diff
@@ -257,6 +257,27 @@ const routes = {
         `https://updown.io/api/checks/${slug}/downtimes`,
     },
   },
+  dRoles: {
+    base: 'https://roles.preprod.data.gouv.fr',
```
XavierJp, line 261: Not sure `base` gets used

models/d-roles/index.ts
XavierJp, line 1: `models/d-roles/index.ts` does not seem like a relevant name / model. I would prefer `groups` as this is all about group membership

This is actually where I would use classes rather than in clients. Maybe a User and a Groups class. 

You raise an HttpUnauthorized but at this point it is not about authorisation,only about groups. 

This mixes several levels of abstration : 
- `models/authentication/agent/agent-connected/index.tsx` is handling authorization
- `models/d-roles/index.ts` is handling group membership. I t CAN raise a HttpUnauthorized only when relevant

Error raised at this level should be requalified at `models/authentication/agent/agent-connected/index.tsx` level.

You have a lot of redundant `try {} catch {} + logError` that would benefit from some factorization.

models/d-roles/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    return await droleApiClient.getGroupsByEmail(email);
+  } catch (error) {
+    logFatalErrorInSentry(
+      new FetchRessourceException({
+        ressource: 'D-Roles',
+        cause: error,
+      })
+    );
+    throw new HttpUnauthorizedError('Failed to get groups by email');
```
XavierJp, line 23: Does not sound like a `HttpUnauthorizedError`

models/d-roles/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    return await droleApiClient.getRoles();
+  } catch (error) {
+    logFatalErrorInSentry(
+      new FetchRessourceException({
+        ressource: 'D-Roles',
+        cause: error,
+      })
+    );
+    throw new HttpUnauthorizedError('Failed to get roles');
```
XavierJp, line 37: Same remark

models/d-roles/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    return await droleApiClient.getUserByEmail(email);
+  } catch (error) {
+    logFatalErrorInSentry(
+      new FetchRessourceException({
+        ressource: 'D-Roles',
+        cause: error,
+      })
+    );
+    throw new HttpUnauthorizedError('Failed to get user by email');
```
XavierJp, line 51: Idem

XavierJp (changes requested):

XavierJp (approved):
Very nice

I think we need to rethink how it will behave in production. 

I would advise to have a temporary variable `D_ROLES_ENABLED` that need to be set to `enabled` in order to use d-roles in prod. However, the way we instantiate `droleApiClient` at the end of `clients/api-d-roles/client.ts`, means that when we import it will trigger its constructor and the 'D-Roles env variables are undefined' error will be thrown.

We can either populate env var with dummy, or we can add an additionnal check on `D_ROLES_ENABLED` that would silence the 'D-Roles env variables are undefined' error.

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const removeUserFromGroup = async (
+  groupId: number,
+  userId: number
+): Promise<void> => {
+  const route = routes.dRoles.groups.removeUserFromGroup(groupId, userId);
+
+  await droleApiClient.fetch<null>(route, {
+    method: 'DELETE',
+  });
```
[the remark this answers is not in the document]
rmonnier9: As there is no return statement it is void.
But indeed for consistency with other functions, I add a return and set the return type to null.

clients/api-d-roles/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  try {
+    const route = routes.dRoles.groups.getGroupsByEmail(email);
+    return await droleApiClient.fetch<IDRolesGroupSearchResponse>(route, {
+      method: 'GET',
+    });
+  } catch (error) {
+    if (error instanceof HttpUnauthorizedError) {
+      return [];
+    }
+    throw error;
```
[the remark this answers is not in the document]
rmonnier9: I guess it was for testing purpose but don't need it anymore.

clients/api-d-roles/client.ts
```diff
@@ -0,0 +1,95 @@
...
+    } catch (e) {
+      this._token = null;
+      throw new HttpUnauthorizedError('Failed to get token');
+    }
+  };
+
+  private isTokenExpired = () => {
+    const now = new Date().getTime();
+    const tokenExpiryTime = this._token ? this._token.expires_in : 0;
+    return now > tokenExpiryTime;
```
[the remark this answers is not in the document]
rmonnier9: Good point, i multiply by 1000 as the other is milliseconds

models/groups/index.ts
```diff
@@ -0,0 +1,136 @@
+import droleClient from '#clients/api-d-roles';
+import {
+  IDRolesGroup,
+  IDRolesGroupSearchResponse,
+} from '#clients/api-d-roles/interface';
+import { HttpUnauthorizedError } from '#clients/exceptions';
+import { FetchRessourceException } from '#models/exceptions';
+import { logFatalErrorInSentry } from '#utils/sentry';
+
+export class GroupModel {
```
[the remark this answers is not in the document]
rmonnier9: I've made the change, good point

models/groups/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Get all groups that a user belongs to
+   */
+  async getGroupsByEmail(email: string): Promise<IDRolesGroupSearchResponse> {
```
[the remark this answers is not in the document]
rmonnier9: Good idea, done !

models/authentication/agent/agent-connected/index.tsx
```diff
@@ -92,7 +93,20 @@ export class AgentConnected {
...
+    const superAgentScopesFromS3 = await superAgentsList.getScopeForAgent(
+      this.email
+    );
+
+    // TEMP Combine scopes from both D-Roles API and S3 storage
+    // Remove any duplicate scopes using Set
+    const superAgentScopes = [
+      ...new Set(superAgentScopesFromDroles),
+      ...new Set(superAgentScopesFromS3),
+    ];
```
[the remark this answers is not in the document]
rmonnier9: Good idea, done

models/authentication/agent/scopes/index.ts
```diff
@@ -22,3 +27,28 @@ export const allAgentScopes = [
...
+    const drolesGroups = await groupModel.getGroupsByEmail(email);
+
+    const allScopesWithDuplicates = drolesGroups
+      .map((group) => {
+        const { validScopes } = parseAgentScope(group.scopes);
+        return validScopes;
+      })
+      .flat();
+
+    const uniqueScopes = [...new Set(allScopesWithDuplicates)];
```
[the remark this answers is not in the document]
rmonnier9: DOne

rmonnier9:
> Very nice
> 
> I think we need to rethink how it will behave in production.
> 
> I would advise to have a temporary variable `D_ROLES_ENABLED` that need to be set to `enabled` in order to use d-roles in prod. However, the way we instantiate `droleApiClient` at the end of `clients/api-d-roles/client.ts`, means that when we import it will trigger its constructor and the 'D-Roles env variables are undefined' error will be thrown.
> 
> We can either populate env var with dummy, or we can add an additionnal check on `D_ROLES_ENABLED` that would silence the 'D-Roles env variables are undefined' error.

Done.

I think we are almost ready.
I will still do some tests.

models/authentication/agent/scopes/index.ts
```diff
@@ -1,24 +1,58 @@
...
+
+    for (const input of scopeArrays) {
+      if (!input) continue;
+
+      if (typeof input === 'string') {
+        const { validScopes } = parseAgentScope(input);
+        allScopes.push(...validScopes);
+      } else if (Array.isArray(input)) {
+        allScopes.push(...input);
+      }
```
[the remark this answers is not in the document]
XavierJp: I advise to only use string[] as input to simplify constructor, and rather ensure that `superAgentScopesRaw` and `groupScopesRaw` are both string

models/groups/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    try {
+      return await droleClient.getGroupsByEmail(email);
+    } catch (error) {
+      logFatalErrorInSentry(
+        new FetchRessourceException({
+          ressource: 'D-Roles Groups',
+          cause: error,
+        })
+      );
+      return [];
```
XavierJp, line 25: Thinking. Should we return an empty list or completly fail the process. An empty would be very frustrating for a user that knows he belongs to some groups

models/groups/index.ts
```diff
@@ -0,0 +1,77 @@
...
+
+      const allScopesWithDuplicates = drolesGroups
+        .map((group) => {
+          const { validScopes } = parseAgentScope(group.scopes);
+          return validScopes;
+        })
+        .flat();
+
+      const uniqueScopes = [...new Set(allScopesWithDuplicates)];
+
```
XavierJp, line 44: This is not needed anymore, it can and should be handled by the `Scope` constructor in `models/authentication/agent/scopes/index.ts` idem for SuperAgentList

models/groups/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    } catch (error) {
+      logFatalErrorInSentry(
+        new FetchRessourceException({
+          ressource: 'D-Roles Admin Groups',
+          cause: error,
+        })
+      );
+      return [];
+    }
+  }
```
XavierJp, line 76: As this is never used, should we remove it ? Less code, less tests

XavierJp (approved):
Better and better

models/groups/index.ts
rmonnier9: Indeed !

models/groups/index.ts
rmonnier9: Indeed, love it

clients/api-d-roles/client.ts
```diff
@@ -0,0 +1,103 @@
...
+class DRolesAPIClient {
+  private _accessToken: string | null;
+  private _tokenExpiryTime: number;
+
+  constructor(
+    private client_id: string | undefined,
+    private client_secret: string | undefined
+  ) {
+    if (
+      process.env.D_ROLES_ENABLED === 'true' &&
```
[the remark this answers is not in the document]
rmonnier9: Done !

models/authentication/agent/scopes/index.ts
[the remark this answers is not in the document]
rmonnier9: Better indeed

models/groups/index.ts
rmonnier9: For login, it will result in a login failure.
Also, for action that requires admin rights, it will also fail.

I both case i think this is the wanted behaviour, i've made the change

models/group/index.ts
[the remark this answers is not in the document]
rmonnier9: Yes

models/authentication/agent/scopes/index.ts
```diff
@@ -1,24 +1,57 @@
...
+
+    for (const input of scopeArrays) {
+      if (!input) continue;
+
+      const scopeStrings = Array.isArray(input) ? input : [input];
+      for (const scopeString of scopeStrings) {
+        const { validScopes } = parseAgentScope(scopeString);
+        allScopes.push(...validScopes);
+      }
+    }
```
XavierJp, line 18: This is hard to read

models/group/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    } catch (error) {
+      logFatalErrorInSentry(
+        new FetchRessourceException({
+          ressource: 'D-Roles Group Admin Check',
+          cause: error,
+        })
+      );
+      throw error;
+    }
+  }
```
XavierJp, line 45: Overkill to call Groups.find then filter it by Id, but let s focus on this later

XavierJp (approved):

models/groups/index.ts
```diff
@@ -30,7 +30,7 @@ export class Groups {
     try {
       const drolesGroups = await this.find(email);
 
-      const scopes = new Scopes(...drolesGroups.map((group) => group.scopes));
+      const scopes = new Scopes(drolesGroups.map((group) => group.scopes));
```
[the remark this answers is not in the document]
rmonnier9: doing this, we change the return type :
Promise<IAgentScope[]>.  --> Promise<string>

It is better to have the first no ?

For now, we have to call Scopes again because of the double use of S3 and D-Roles, later we will remove it

models/authentication/agent/scopes/index.ts
```diff
@@ -2,24 +2,40 @@ import { IAgentScope } from './constants';
...
-      const scopeStrings = Array.isArray(input) ? input : [input];
-      for (const scopeString of scopeStrings) {
-        const { validScopes } = parseAgentScope(scopeString);
-        allScopes.push(...validScopes);
-      }
+    const scopeStrings = Array.isArray(scopes) ? scopes : [scopes];
+    for (const scopeString of scopeStrings) {
+      const { validScopes } = parseAgentScope(scopeString);
+      allScopes.push(...validScopes);
     }
```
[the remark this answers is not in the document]
rmonnier9: yes, so we only use the method add

PR #1860, opened 2025-07-09
fix: allow fallback to S3 in Groups class without throwing error and …

rmonnier9:
- Changement mineur.
- Détails :
  - Fallback sur S3 sans erreur si d-roles renvoie une erreur
  - Log sur Sentry des différences entre d-roles et S3

related to #1854

### DIFF
--- a/models/authentication/agent/agent-connected/index.tsx
+++ b/models/authentication/agent/agent-connected/index.tsx
@@ -5,8 +5,10 @@ import {
   NeedASiretException,
   PrestataireException,
 } from '#models/authentication/authentication-exceptions';
+import { InternalError } from '#models/exceptions';
 import { Groups } from '#models/groups';
 import { isSiret, verifySiret } from '#utils/helpers';
+import { logWarningInSentry } from '#utils/sentry';
 import { AgentOrganisation } from '../organisation';
 import { defaultAgentScopes } from '../scopes/constants';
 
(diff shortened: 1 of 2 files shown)


models/authentication/agent/agent-connected/index.tsx
```diff
@@ -95,11 +97,14 @@ export class AgentConnected {
...
   async getAgentHabilitation() {
     const superAgentScopes = new Scopes();
+    let drolesScopes: string[] = [];
+    let s3Scopes: string[] = [];
 
     if (process.env.D_ROLES_ENABLED === 'enabled') {
       // Get scopes from Groups (D-Roles API) for this agent
       const groups = await Groups.find(this.email, this.userId);
       groups.forEach((group) => {
+        drolesScopes.push(...group.scopes);
```
XavierJp, line 107: Why not use a `Set` here ?

models/authentication/agent/agent-connected/index.tsx
```diff
@@ -108,8 +113,37 @@ export class AgentConnected {
...
 
+    // Compare scopes from D-Roles and S3, log differences to Sentry
+    if (process.env.D_ROLES_ENABLED === 'enabled' && drolesScopes.length > 0) {
+      const drolesScopesSet = new Set(drolesScopes);
+      const s3ScopesSet = new Set(s3Scopes);
+
+      const onlyInDroles = drolesScopes.filter(
+        (scope) => !s3ScopesSet.has(scope)
+      );
+      const onlyInS3 = s3Scopes.filter((scope) => !drolesScopesSet.has(scope));
```
XavierJp, line 127: Simpler way to compare them could be https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/union

Union both set and compare lenght of union with length of any of the two sets

XavierJp: But you would have the `onlyInDroles` and `onlyInS3` variables

XavierJp (approved):
So this PR is only about adding logs to D-Roles / S3 logic ?

PR #1949, opened 2025-08-26
feat: first implementation of activation link redirect

XavierJp:
First implementation, rely heavily on exception bubbling. Not the best as it will completely prevent user from logging in

image

But it means that if a user cannot activate its account, it can no longer use EspaceAgent, even id the issue is on roles.data or ProConnect side. 

Improvement over this design would be to to just display a banner and still allow agent to connect.


HAEKADI (approved):

XavierJp:
EN V2, je n'affiche le warning que dans la page recap des droits 
image

PR #2002, opened 2025-09-17
fix: keep query params when redirect

rmonnier9:
- Correction d'un bug. 
- Détails :
  - Préservation des paramètres d'URL lors de la redirection vers la page canonique
  - Revert robots.txt that doesn't index redirected=1

### DIFF
--- a/cypress/e2e/recherche/research-redirections.cy.js
+++ b/cypress/e2e/recherche/research-redirections.cy.js
@@ -8,7 +8,7 @@ describe('Siren / Siret redirections', () => {
 
     cy.get('.fr-search-bar > button').click();
 
-    cy.url().should('include', '/entreprise/danone-552032534');
+    cy.url().should('include', '/entreprise/danone-552032534?redirected=1');
   });
 
   it('Unformatted siren redirection', () => {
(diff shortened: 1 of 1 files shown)


XavierJp (changes requested):
I am a bit surprised. Are you sure it used to be done this way ?

Can you please git blame to be sure

XavierJp (changes requested):
I would entirely remove line 68-80

app/(header-default)/entreprise/[slug]/page.tsx
```diff
@@ -66,12 +66,21 @@ export default async function UniteLegalePage(props: AppRouterProps) {
   ]);
 
   const extractedSiren = extractSirenOrSiretSlugFromUrl(slug);
+  // We redirect from /entreprise/${siren} to /entreprise/${slug}
+  // Nb: in somes cases, there can be a two redirects :
+  // /rechercher?terme=${siren} -> /entreprise/${siren}?isRedirected=1 -> /entreprise/${slug}?isRedirected=1
```
Comment (rmonnier9) [inline app/(header-default)/entreprise/[slug]/page.tsx:71]: Indeed this is better like this !

Just a little change :

```suggestion
  // /rechercher?terme=${siren} -> /entreprise/${siren}?redirected=1 -> /entreprise/${slug}?redirected=1
```

PR #2091, opened 2025-10-22
[ExportSirene]fix(csv): improves export error messages

AlanChauchet:
related to #2089


HAEKADI (approved):
Great !
@XavierJp For some context https://github.com/annuaire-entreprises-data-gouv-fr/site/issues/2089
The error handling is not very explicit. So if your file has more that the allowed 500 values. You simply get a `Validation error` with no explanation.

PR #2096, opened 2025-10-23
feat/Use resource server instead of OAuth for groups operations

XavierJp:
- [x] Remove all not verified logic
- [x] Add resource server routes in ProConnect
- [x] Ensure XSS attacks protection is top notch
- [x] Remove all OAuth group logic
- [x] Implement refresh token logic in ProConnect
- [x] Migrate entirely from page router and clean legacy code
- [x] Implement error handling and redirection, especially when proConnect token expired
- [x] admin check front/back/roles need to be challenged with @AlanChauchet 
- [x] Remove client_id and client_secret from repo rename every D_ROLES_ into ROLES_DATA

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -1,7 +1,7 @@
 import { type SentryBuildOptions, withSentryConfig } from "@sentry/nextjs";
 import type { NextConfig } from "next";
 import getContentSecurityPolicy from "#utils/headers/content-security-policy";
-import { getBaseUrl } from "#utils/server-side-helper/app/get-base-url";
+import { getBaseUrl } from "#utils/server-side-helper/get-base-url";
 import redirects from "./redirects.json" with { type: "json" };
 
 const WITH_SENTRY =
--- a/app/(header-compte)/compte/accueil/page.tsx
+++ b/app/(header-compte)/compte/accueil/page.tsx
@@ -2,6 +2,7 @@ import type { Metadata } from "next";
 import { redirect } from "next/navigation";
 import AgentNavigation from "#components/espace-agent-components/agent-navigation";
 import { FullTable } from "#components/table/full";
+
 import { Badge } from "#components-ui/badge";
 import ButtonLink from "#components-ui/button";
 import FullWidthContainer from "#components-ui/container";


XavierJp:
NB : going through every MatomoEvent made me questionned some stuff we logged an could stop logging @HAEKADI 
 
Especially I removed an old log on TVA, that use to log real 200 from fake 200

app/api/auth/agent-connect/login/route.ts
```diff
@@ -2,15 +2,18 @@ import { NextResponse } from "next/server";
...
+} from "#utils/session";
 import withSession from "#utils/session/with-session";
 
 export const GET = withSession(async function loginRoute(req) {
   try {
-    if (req.session.franceConnectHidePersonalDataSession) {
-      throw new Error("User is already logged in with FranceConnect");
-    }
+    await cleanAgentSession(req.session);
+    await cleanFranceConnectSession(req.session);
```
AlanChauchet: Neat 💄 So now we can be connected to FranceConnect and silently be logged out + logged into ProConnect seamlessly ?

models/authentication/agent/agent-connected/index.tsx
```diff
@@ -103,24 +94,23 @@ export class AgentConnected {
 
     return {
       ...habilitationInheritedFromOrganisation,
-      agentIsNotVerified,
     };
```
AlanChauchet:
```suggestion
return habilitationInheritedFromOrganisation;
```
if we don't need to clone the object

next.config.ts
```diff
@@ -53,7 +52,17 @@ const nextConfig: NextConfig = {
...
           {
             key: "Content-Security-Policy",
-            value: getContentSecurityPolicy(),
+            value: [
+              "default-src 'self'",
+              "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://stats.data.gouv.fr/",
+              "style-src 'self' 'unsafe-inline'",
+              "img-src 'self' data:",
+              "frame-src 'self' https://stats.data.gouv.fr/ https://plugins.crisp.chat/",
+              "connect-src 'self' https://stats.data.gouv.fr/ https://errors.data.gouv.fr/ https://bodacc-datadila.opendatasoft.com/ https://data.economie.gouv.fr/ https://journal-officiel-datadila.opendatasoft.com/ https://api-lannuaire.service-public.fr/ https://data.culture.gouv.fr/ https://data.inpi.fr/ https://openmaptiles.geo.data.gouv.fr/ https://openmaptiles.data.gouv.fr/ https://geo.api.gouv.fr https://api-adresse.data.gouv.fr https://tabular-api.data.gouv.fr https://koumoul.com",
```
AlanChauchet: This limits the endpoints from which we can use `fetch`, right ? 👀 (but only from the client since it's a response header)

PR #2121, opened 2025-10-30
📥[ExportSirene]feat: adds naf codes file upload

AlanChauchet:
related to #2083


HAEKADI (changes requested):
Same as the location PR, I feel like we can guide this more in terms of file content format:

Screenshot 2025-11-07 at 17 55 06

What do you think?

[ref](https://sirene.fr/sirene/public/verifier-creation-fichier.action#activite-principale)

PR #2193, opened 2025-12-11
fix(sentry): removes user is a bot error from sentry

AlanChauchet:
"User is a bot" error is poping way too much in Sentry, with almost 1 million occurrences in a week


AlanChauchet:
> @PaulLesur prefers we keep it for now, until the infra is stable.

@HAEKADI okay but we might soon reach our errors quota. I'll check regularly to see if it's the case 😬

PR #2206, opened 2025-12-16
feat(espace agent): remove nj 7321 from list of might be authorised admin

HAEKADI:
Depends on https://github.com/annuaire-entreprises-data-gouv-fr/search-infra/pull/631
Nature juridique 7321 has been added to the L100_3 list. So it can access l'espace agent without an issue.


HAEKADI:
updated in #2527

PR #2241, opened 2026-01-20
fix(etablissements): adds noindex, nofollow for closed etablissements

AlanChauchet:


HAEKADI (approved):
lgtm!
Let's hold off for a few weeks until Google indexation is more stable and we'll revisit.

PR #2396, opened 2026-04-15
feat(build): build and push image

PaulLesur:
- **feat(build): add Dockerfile**
- **feat(build): add build step**


.github/workflows/deploy.yml
```diff
@@ -8,6 +8,83 @@ on:
...
+          INSEE_PASSWORD_AGENTS=${{ secrets.INSEE_PASSWORD_AGENTS }}
+          INDEXING_ENABLED=enabled
+          MATOMO_ENABLED=disabled
+          VITE_MATOMO_SITE_ID=${{ secrets.MATOMO_SITE_ID }}
+          MATOMO_API_SITE_ID=${{ secrets.MATOMO_API_SITE_ID }}
+          MATOMO_TOKEN_AUTH=${{ secrets.MATOMO_TOKEN_AUTH }}
+          VITE_IRON_SESSION_PWD=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+          GRIST_API_KEY=${{ secrets.GRIST_API_KEY }}
+          UPDOWN_IO_API_KEY=${{ secrets.UPDOWN_IO_API_KEY }}
+          EOF
```
AlanChauchet: For the build we only need `VITE_*` env vars. The other vars are not necessary until the server is running.
But here we're missing `VITE_DATAPASS_URL`, `VITE_SENTRY_DSN` and `VITE_BASE_URL` (`VITE_SENTRY_DSN` and `VITE_BASE_URL` differs between staging and prod). Also `VITE_IRON_SESSION_PWD` does not exist

PaulLesur: > differs between staging and prod

Do you mean that we still have to make a different build for staging and production ?

PaulLesur: For the env vars, I took them from here : https://github.com/annuaire-entreprises-data-gouv-fr/site/blob/main/.github/workflows/pre-merge.yml

Maybe we need to update this part.

AlanChauchet: > Do you mean that we still have to make a different build for staging and production ?

Yes. The "frontend env vars" are not real env vars, they're read at build time and inlined in the client generated code. So in order to make them differ between staging and prod, we still need to have different builds.

PaulLesur: OK I dissociated the two builds.

.github/workflows/deploy.yml
```diff
@@ -55,7 +148,7 @@ jobs:
           - environment: production-03
             host: SSH_PRODUCTION_03
     steps:
-      - uses: actions/checkout@v6
```
AlanChauchet, line 58: why the change of checkout version ?
we get some warnings with v4

AlanChauchet (approved):

PR #2628, opened 2026-07-16
feat(banner): adds roles migration banner

AlanChauchet:


HAEKADI (approved):
We have to wait for their go before changing the time!

